PR #1, opened 2022-03-10
BDR-275: Add `template.json` to the MVP DWC Template

HaydenGaia:
## Description
* This PR adds the discussed `template.json` to the template skeleton
* Questions to resolve before merging:
  * Should the filename be `template.json`, or `01_dwc_mvp.json`?
  * Should `id` be `01_dwc_mvp` or `01_dwc_mvp.xlsx` (i.e., include extension)?
  * If the `id` is not the filename, should we include a `"filename": "01_dwc_mvp.xlsx"`?
  * Should we include `instructions` at the moment, when none exist?

### DIFF
--- a/templates/01_dwc_mvp/metadata.json
+++ b/templates/01_dwc_mvp/metadata.json
@@ -1,0 +1,5 @@
+{
+    "id": "01_dwc_mvp.xlsx",
+    "description": "A template to translate some Darwin Core fields",
+    "version": "0.0.1"
+}
--- a/templates/01_dwc_mvp/template.json
+++ b/templates/01_dwc_mvp/template.json
@@ -1,6 +1,0 @@
-{
-    "id": "01_dwc_mvp",
-    "description": "A template to translate some Darwin Core fields",
-    "version": "0.0.1",
-    "instructions": "See README.md for further details"
-}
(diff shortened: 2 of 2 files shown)


serge-gaia:
* Should the filename be template.json, or 01_dwc_mvp.json?: **template.json**
My initial idea was to use 01_dwc_mvp.json (templateId.json) but I thought that it might confuse people between  `01_dwc_mvp.xlsx` and  `01_dwc_mvp.json`, which one is the real template. Also if we write a software that looks up all the templates we just need to look for the `template.json` files without knowing the id of the templates beforehand (I know we can lookup the subdir but still). So let's keep it with a common name to all templates. It's just the name itself that I don't really like (template.json). I had thought about `template_description.json`, `settings.json`, ... I'm welcoming any naming suggestion

* id with extension or not?: **yes**
Initially I was thinking no extensions in the id, because I was imagining the case where we have an xlsx and a csv for example for the same 'template' but I now I think that a 'template' is actually a single file. **So yes to extension in the id.**

* should we include instructions at the moment, when none exist? : no
no need of instructions in the json for now. I believe that ultimately the instructions are going to be html pages.

serge-gaia:
@HaydenGaia , reassigning to you to incorporate my comments above

HaydenGaia:
@serge-gaia re: the filename, how about `metadata.json`?

serge-gaia:
`metadata.json` it is!

HaydenGaia:
@serge-gaia Realised I can't re-assign to you, but ready to be reviewed again :+1:

PR #2, opened 2022-03-14
BDR-415: Project Restructure, Base Class & Poetry-ification

SupImDos:
## Description
* This PR contains:
  * Project restructure:
    * Renamed `01_dwc_mvp` to `dwc_mvp` (removed leading digits)
    * `templates` now resides inside the `abis_mapping` directory
    * This allows templates to and mappers to be accessed and packaged properly by poetry
  * Poetry-ification of project:
    * Project can be built and installed as a regular python package
    ```python
    import abis_mapping
    mapper = abis_mapping.utils.get_mapper("dwc_mvp")
    ...
    ```
  * `ABISMapper` base class and package interface:
    * `ABISMapper` provides a base class for all template mappers to inherit from
    * Concrete `ABISMapper`s must register their template ID with the base class
    * `loader` dynamically loads all of the template mappers
    * `utils` provides interface to retrieve mappers based on template ID


abis_mapping/base.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class ABISMapper(abc.ABC):
+    """ABIS Mapper Base Class"""
+
+    # Concrete ABIS Mapper Registry
+    registry: dict[str, type["ABISMapper"]] = {}
+
+    @abc.abstractmethod
+    def apply_mapping(self, data: bytes) -> rdflib.Graph:
```
serge-gaia, line 21: I think the input data type should be at higher level than `bytes`, something like StringIO that support buffering and streaming with the decoding (UTF-8) done by the client.
If we decide to use pandas for our CSV reader, and I think we should, let's set the data type the same as `pandas.read_csv()` function: 
https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html
(`FilePath | ReadCsvBuffer[bytes] | ReadCsvBuffer[str]`)

abis_mapping/utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+    Args:
+        template_id (str): Template ID to retrieve the mapper for
+
+    Returns:
+        type[ABISMapper]: ABIS mapper associated with the template ID.
+
+    Raises:
+        KeyError: Raised if the template ID supplied cannot be found
+    """
```
serge-gaia, line 19: I'd rather see the function returning `None` instead of raising an exception.

abis_mapping/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    # Retrieve and return the mapper
+    return base.ABISMapper.registry[template_id]
```
serge-gaia, line 21: to return `None` use the `get()` method of dict.
return base.ABISMapper.registry.get(template_id)

abis_mapping/utils.py
SupImDos: Resolved in: ff096ef

abis_mapping/utils.py
SupImDos: Resolved in: ff096ef

abis_mapping/base.py
SupImDos: Added the type hinting as above in: eb060b8

Interestingly, I could not get my IDE to resolve the `pandas._typing.FilePath` / `pandas._typing.ReadCsvBuffer` imports, so not sure if this will be a problem.

PR #5, opened 2022-03-22
BDR-389: Re-define `pandas` types so they can be resolved by `mypy`

HaydenGaia:
## Description
* Importing types from `pandas._typing` does not work (for whatever reason, I cannot find out)
* `mypy` complains that the `types.CSVType` is of type `Union[Any, Any, Any]`
* This PR re-defines the pandas types, so they can be resolved properly


serge-gaia:
strange this mypy issue.

PR #7, opened 2022-03-28
BDR-285: Frictionless Data Validation

HaydenGaia:
## Description
This PR adds:
* Fully migrated from `pandas` to `frictionless`
* Added `utils.checks` to provide custom checks for frictionless validation
  * `ValidateCoordinates`: validates the supplied lat/long with `utils.coords`
  * `NotEmpty`: validates that the supplied table is not empty (0 rows)
  * `NotTabular`: validates that the raw data supplied is tabular in nature
* Added `utils.coords` to provide coordinate validation
  * Implements coordinate ranges from Rocio in BDR-397
  * Provides method to check if lat/long are within ranges
* Added `base.exceptions` base exceptions
  * `ABISMapperError`
  * `ABISMapperMappingError`
  * `ABISMapperValidationError`
* Added abstract method `apply_validation()` onto base `ABISMapper` class
  * Receives readable raw data, validates and returns `frictionless.Resource`
  * If invalid, raises exception which includes frictionless validation report
* Added class method `metadata()` onto base class, which deserialises and caches the `metadata.json` for a template
* Added class method `schema()` onto base class, which deserialises and caches the `schema.json` for a template
* Added `apply_validation()` for `dwc_mvp.xlsx` template
  * Validates using `schema.json`
  * Validates with extra checks: `ValidateCoordinates`, `NotEmpty`, `NotTabular`
* Added initial `schema.json` for `dwc_mvp.xlsx` template
* Improved template registry, with template id as class attribute on template mapper class
* Changed abstract method `apply_mapping()` to receive `frictionless.Resource` as input
* Fixed example output `ttl` (previously had precision truncated, as parsed as `float` instead of `Decimal`) (good!)
* Simplified `dwc_mvp.xlsx` template by making all timestamps "dates" (for now)
* Improved `dwc_mvp.xlsx` template mapper by removing references to `pandas`
* Improved and genericised `base.types`, removing references to `pandas`

## Complete Success Criteria
* > Incorrect format (format must be csv or geopackage, note that xlsx is not accepted)
  * Currently the "infrastructure" places no restrictions on file format - this is left to the implementation of each template `mapping.py`.
  * Currently, the only existing template (`dwc_mvp.xlsx`) accepts both `csv` and `xlsx`.
  * Including `xlsx` cost nothing (built in), can remove if we really want though.
  * Validator includes extra check: `utils.checks.NotTabular`
  * Validator detects if the data supplied is at least tabular in nature
  * If it is not tabular (e.g., `csv`, `xlsx`), marks as error and adds to report
* > Missing mandatory fields
  * Validator has `required` constraints on fields in `schema.json`
  * Validator detects missing mandatory fields, marks as an error and adds to report
* > Blank template
  * Validator includes extra check: `utils.checks.NotEmpty`
  * Validator detects if template is blank, marks as an error and adds to report
* > Empty rows
  * Validator has inbuilt `blank-row` checker
  * Validator detects if a row is blank, marks as an error and adds to report
* > Schema errors
  * Validator uses `schema.json` to validate
  * If it does not match schema, marks as error and adds to report
* > Latitude and longitude not between specified min & max values (see Rocio's comment on BDR-397 for acceptable ranges)
  * Validator includes exyta check: `utils.checks.ValidCoordinates`
  * Validator uses coordinate ranges from BDR-397 to check that latitude and longitude are valid
  * If the latitude and longitude do not fall into a valid range, marks as error and adds to report

## Incomplete Success Criteria
* Integration with BDR Part-A API Endpoint


abis_mapping/templates/dwc_mvp/schema.json
```diff
@@ -0,0 +1,125 @@
...
+    },
+    {
+      "name": "decimalLatitude",
+      "title": "Decimal Latitude",
+      "description": "...",
+      "example": "-33.80",
+      "type": "number",
+      "format": "default",
+      "constraints": {
+        "required": true
```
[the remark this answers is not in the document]
HaydenGaia: Fixed in: d6d5bf6

abis_mapping/templates/dwc_mvp/schema.json
```diff
@@ -0,0 +1,125 @@
...
+    {
+      "name": "decimalLongitude",
+      "title": "Decimal Longitude",
+      "description": "...",
+      "example": "115.21",
+      "type": "number",
+      "format": "default",
+      "constraints": {
+        "required": true
+      }
```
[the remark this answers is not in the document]
HaydenGaia: Fixed in: d6d5bf6

HaydenGaia:
### Update
* Refactored interface such that the expected usage is below.
* This removes strange `try/catch` interface that was required before (unnecessary)
```python
# Get Mapper
mapper = abis_mapping.get_mapper(template_id)

# Check if mapper exists
if not mapper:
    print(f"The mapper '{template_id}' isn't real!")
    return

# Apply Validation (Get Validation Report)
valid = mapper().apply_validation(raw_data)

# Check Report
if not report.valid:
    print(f"Your data is invalid, here is the report: {report}")
    return

# Apply Mapping
graph = mapper().apply_mapping(raw_data)  # Data is already valid here!

# Serialize
ttl = graph.serialize(format="ttl")
```

abis_mapping/utils/coords.py
```diff
@@ -0,0 +1,111 @@
...
+    # Coral Sea Islands
+    CoordinateRange(
+        decimal.Decimal("-29.982747999620415"),
+        decimal.Decimal("147.83945691360577"),
+        decimal.Decimal("-15.721024000057525"),
+        decimal.Decimal("159.1407290000434"),
+    ),
+    # Australian Antarctic Territory
+    # TODO -> TBC (BDR-397)
+]
```
serge-gaia, line 86: I think it should be a dict with named coordinates instead of a list but there's a lot to discuss about that. Especially these crazy-nonsense-15-digits-after-decimal-place coordinates. They will required the client to have 15 digits GPS precision, which doesn't exist

abis_mapping/utils/checks.py
```diff
@@ -0,0 +1,109 @@
...
+class ValidCoordinates(frictionless.Check):
+    """Checks whether the latitude and longitude of the row are valid."""
+
+    # Check Attributes
+    code = "valid-coordinates"
+    Errors = [frictionless.errors.RowConstraintError]
+
+    def __init__(
+        self,
+        descriptor: Any=None,
```
serge-gaia, line 23: PEP 8: E252 missing whitespace around parameter equals

abis_mapping/utils/coords.py
serge-gaia: don't worry about it now. There's much to discuss about these 'local' validations rules. I think that in this repo we should stick with basic frictionless validation rules and offer the possibility for the client to pass its own validator in `ABISMapper.apply_validation(...)`

abis_mapping/base/mapper.py
```diff
@@ -20,35 +25,82 @@ class ABISMapper(abc.ABC):
...
     registry: dict[str, type["ABISMapper"]] = {}
 
+    # ABIS Mapper Template ID
+    template_id: str = NotImplemented  # Must be implemented
+
+    @abc.abstractmethod
+    def apply_validation(
+        self,
+        data: types.ReadableType,
+        ) -> frictionless.Report:
```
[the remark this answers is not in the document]
HaydenGaia: Fixed in: 4cf2459

abis_mapping/utils/checks.py
HaydenGaia: Fixed in: 4cf2459

abis_mapping/utils/coords.py
HaydenGaia: Truncated coordinates to 6 d.p. in: 883ca70
As discussed, leaving coordinate checks in for now.
We can remove/refactor them into BDR Part-A later if required

HaydenGaia:
@serge-gaia Ready for re-review

PR #24, opened 2022-05-30
BDR-621: Initial pass of lazy iterator chunking interface

HaydenGaia:
## Description
* Changes `ABISMapper.apply_mapping()` interface to return an `Iterator[rdflib.Graph]` instead of just `rdflib.Graph`
* Added argument `chunk_size: Optional[int]` to `ABISMapper.apply_mapping()` to specify chunk size
* Implementors are expected to provide a lazy iterator interface for mappers
* Added `utils.chunking.should_chunk()` utility function to determine when to yield a chunk

### DIFF
--- a/abis_mapping/templates/dwc_mvp/mapping.py
+++ b/abis_mapping/templates/dwc_mvp/mapping.py
@@ -13,7 +13,7 @@ from abis_mapping import base
 from abis_mapping import utils
 
 # Typing
-from typing import Optional
+from typing import Iterator, Optional
 
 
 # Default Dataset Metadata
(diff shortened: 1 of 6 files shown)


abis_mapping/templates/dwc_mvp/mapping.py
```diff
@@ -105,6 +106,10 @@ def apply_mapping(
             onerror="raise",  # Raise errors, it should already be valid here
         )
 
+        # Infer Statistics and Count Number of Rows
+        resource.infer(stats=True)
```
kehh, line 110: How safe is this `infer` method? Is it part of frictionless?

abis_mapping/utils/chunking.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+# Third-Party
+import frictionless
+
+# Typing
+from typing import Optional
+
+
+def should_chunk(
```
kehh, line 11: Probably need a test for this but I'm sure you've got that on your radar.

kehh (commented):
Some questions and comments

abis_mapping/templates/dwc_mvp/mapping.py
HaydenGaia: @kehh Yes, from `frictionless`: https://framework.frictionlessdata.io/docs/guides/framework/resource-guide/#stats

abis_mapping/utils/chunking.py
HaydenGaia: @kehh Good point, missed them for some reason

HaydenGaia: Unit tests added in 18147e7

PR #27, opened 2022-06-07
BDR-628: Final Mappings for Pilot

HaydenGaia:
#### Mapping
* Added the following fields:
    * `establishmentMeans`
    * `lifeStage`
    * `sex`
    * `reproductiveCondition`
    * `identificationMethod`
    * `associatedSequences`
    * `sequencingMethod`
    * `acceptedNameUsage`
* Updated template, `schema.json` and unit tests to match new field mapping

#### Plugins & Refactoring
* Added new `"list"` custom frictionless plugin
  * Allows for whitespace separated string lists to be parsed and validated from a CSV
  * We will use this instead of the inbuilt `json` encoding (bad user experience?)
* Refactored custom `frictionless` plugins and checks from `utils` to `plugins` directory
* Refactored parts of `types.py` to prevent circular imports caused by new structure
* Refactored unit tests to handle new structure

#### Fixes
* Fixed `conftest.compare_graphs()` unit test function
    * Appears to be a bug with `rdflib.compare.isomorphic` hash calculation
    * Appears blank nodes are handled slightly different between graphs that are constructed programmatically and graphs that are parsed
    * As such, quick fix for unit tests is to do a serialization-deserialization round-trip before comparison


abis_mapping/plugins/list.py
```diff
@@ -0,0 +1,103 @@
...
+        """Convert cell (read direction)
+
+        Args:
+            cell (Any): Cell to convert
+
+        Returns:
+            Optional[list[str]]: Converted cell, or none if invalid.
+        """
+        # Check that cell is not already a "list"
+        if not isinstance(cell, list):
```
serge-gaia, line 63: is this mean that if comma `,` is used as a delimiter frictionless will parse it as a `list`?

abis_mapping/plugins/list.py
```diff
@@ -0,0 +1,103 @@
...
+    constraints = [
+        "required",
+        "minLength",
+        "maxLength",
+        # "pattern",  # TODO -> Check whether this works
+        # "enum",  # TODO -> Check whether this works
+    ]
+
+    # Delimiter
+    delimiter = " "
```
serge-gaia, line 51: mmm, `space` as a delimiter will allow only single word values. I'd rather see a comma as delimiter, unless frictionless already support `,` as delimiter (see comments below)

serge-gaia:
Happy to merge that. I can't say that I have rigorously verify the mapping.

serge-gaia:
let's talk about the list delimiter later.

abis_mapping/plugins/list.py
HaydenGaia: This just means if you construct a cell programmatically you can pass it a `list[T]` without it doing any of the parsing stuff.

This is how its currently done in `frictionless` internally. E.g., for their inbuilt `datetime` type it checks whether the passed in object is a datetime already.

See: https://github.com/frictionlessdata/frictionless-py/blob/main/frictionless/types/datetime.py#L28

abis_mapping/plugins/list.py
HaydenGaia: I agree, there is still a lot of discussion around this `associatedSequences` field that uses this custom list plugin.

It was decided that `json` notation for an array (e.g., `["A", "B", "C"]`) was not user friendly for someone filling out the template.

Originally, this field was meant to only be URIs, so space separator made sense. Then we found out that WA BIO can't guarantee that they will be URIs - so we will have to be able to handle anything.

TLDR: See: BDR-628 comments

kehh:
The delimiters is probably my fault sorry @serge-gaia

PR #29, opened 2022-06-08
BDR-548: Handling Vocabs and Creating Vocabs on the Fly

HaydenGaia:
## Description
* A huge PR to add in proper vocabulary handling and creation of vocabs 'on-the-fly'
* Removed most enum restrictions in `schema.json` (now create vocabs 'on-the-fly')
* Adds in the following new vocab utility classes
  * `vocabs.Term`: Represents a vocab term with 1 or more human readable labels
  * `vocabs.RestrictedVocabulary`: For handling enum-like vocabs where options are restricted
  * `vocabs.FlexibleVocabulary`: For handling vocabs where if the vocab _doesn't_ exist, we create a new one 'on-the-fly'
* Refactors vocabs and terms to new `vocabs` subpackage/directory
  * Each vocab lives in its own file, e.g., `geodetic_datum.py`
* Updated mapping to use new vocabs
* Updated unit tests and test data to match
* Fixed split/join of custom list type fields when multiple spaces were used


kehh:
What happens with terms that we've sent to part B?

HaydenGaia:
> What happens with terms that we've sent to part B? I'm worried that we'll end up creating many duplicates?

We will end up creating duplicates, but we simply have no feedback mechanism from BDR Part-B. Mieke might know at a higher level what the future intention of this is.

HaydenGaia:
> What happens with terms that we've sent to part B? I'm worried that we'll end up creating many duplicates?

I think BDR Part-B will just have to triage and handle the duplicates. It has to handle other duplicate data as is right? Not just concepts.

kehh:
well one day they'll give us a return which includes an iri for the term then we could use that.

PR #32, opened 2022-06-13
BDR-648: set package version to 1.0.0-rc1

serge-gaia:


poetry.lock
```diff
@@ -68,14 +68,14 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
...
-version = "6.3.3"
+version = "6.4.1"
 description = "Code coverage measurement for Python"
 category = "dev"
 optional = false
 python-versions = ">=3.7"
 
 [package.dependencies]
-tomli = {version = "*", optional = true, markers = "extra == \"toml\""}
+tomli = {version = "*", optional = true, markers = "python_full_version <= \"3.11.0a6\" and extra == \"toml\""}
```
serge-gaia, line 78: I did run a `poetry lock --no-update` but it still updated some dependencies. ??

HaydenGaia: Strange, maybe an old version of poetry?
Why do we need to do a `poetry lock` anyway, the lock file is already there?

serge-gaia: Good question. If I update the `.toml` without updating the `.lock` file my IDE gives me a warning. Maybe it's just about timestamps as the `.lock` should always be newer that the `.toml`

serge-gaia: Not to sure actually

HaydenGaia: Hmm, I'm not sure then.
If you want to satisfy the IDE then we can leave it as is (with the dependency updates) as long as unit tests still pass, or we can revert the change.

Up to you.

PR #36, opened 2022-09-05
BDRC-41: Add template skeleton for new `occurrence_extended.csv`

HaydenGaia:
#### Summary
* For `occurrence_extended.csv`, `schema.json` and `margaret_river_flora.csv`
    * Adds new columns as per `BDRC-41` `occurrence_extended_0.4.csv` from Mieke
    * Renames columns as per `BDRC-41` `occurrence_extended_0.4.csv` from Mieke
    * Re-arranges columns as per `BDRC-41` `occurrence_extended_0.4.csv` from Mieke
* For `mapping.py`
    * Updates mapping to match the renamed columns
    * Outputted RDF is the exact same as previous (new columns are not yet mapped)


abis_mapping/templates/occurrence_extended/schema.json
```diff
@@ -261,35 +283,46 @@
...
     },
     {
-      "name": "collectionCode",
-      "title": "Collection Code",
+      "name": "otherCatalogNumbers",
+      "title": "Other Catalog Numbers",
       "description": "...",
-      "example": "ARACH",
-      "type": "string",
+      "example": "BHP2012-7521 | M12378",
```
serge-gaia, line 297: I thought that the list delimiter was a white space. So, is this example with the pipe `|` accurate to show how to make a list?

HaydenGaia: This was the example provided to me by Mieke.

You're right that currently the delimiter is a space, so this will be split to `["BHP2012-7521", "|", "M12378"]`.

But, I believe Mieke wants to change the delimiters in all list fields to a pipe `|`, so that discussion still needs to be had and implemented.

HaydenGaia: Main reason its not comma at the moment is because of the "C" in CSV. I wasn't sure if people would escape the data properly.

Also, one of the list fields is a list of URLs, so it needs to be a character that won't be in a URL (or again, that people will escape properly).

Anyway, yes - we need to discuss with Mieke.

serge-gaia: oh yes the CSV of course.

PR #37, opened 2022-09-08
BDRC-61: Add Flexible Delimiters for List Data Types (Pipe Delimiter)

HaydenGaia:
## Summary
* Adds flexible delimiter to "list" data types
* Allows support of templates with previous/old delimiters (e.g., space `" "`)
* Allows for new templates to use different delimiters (e.g., pipe `"|"` by default)
* Adjusts `occurrence_data.csv` `schema.json` to use `"list[ ]"` space delimiter for perfect backwards compatibility
* Adjusts `occurrence_extended.csv` `schema.json` and example data for new default pipe `"|"` delimiter behaviour
* Adds comprehensive unit testing

## Details
* This PR modifies the `list` field type to include a flexible delimiter parameter as follows:
    * `"list"`: default pipe delimiter
    * `"list[ ]"`: explicit space delimiter
    * `"list[|]"`: explicit pipe delimiter
    * `"list[anything]"`: explicitly provide any delimiter
* This is done using the following regex: `^list(?:\[(.+)\])?$`


HaydenGaia:
As per quick chat, some changes:
* Set pipe `"|"` to the default delimiter
* Change `occurrence_data.csv` `schema.json` to `list[ ]` (space delimited)

PR #38, opened 2022-09-12
BDRC-67: Frictionless Mutual Inclusion Plugin & Custom Checks Refactor

HaydenGaia:
## Summary
* Adds `plugins.mutual_inclusion` with the `MutuallyInclusive` Frictionless checker
    * This checker allows the user to specify fields that _must_ be provided together
* Adds the `MutuallyInclusive` checker to `occurrence_extended.csv` for `threatStatus` and `conservationJurisdiction`
     * As per Mieke in BDRC-42 / BDRC-67, these fields must be provided together or not at all
* Refactors the `plugins.checks` module to (moves checkers into their own modules):
    * `plugin.coordinates` - for coordinate checking plugin
    * `plugin.empty` - empty dataset checking plugin
    * `plugin.tabular` - for tabular dataset checking plugin
* Adds and refactors unit tests for checking plugins

### DIFF
--- a/abis_mapping/plugins/mutual_inclusion.py
+++ b/abis_mapping/plugins/mutual_inclusion.py
@@ -1,0 +1,58 @@
+"""Provides extra frictionless mutual inclusion checks for the package"""
+
+
+# Third-Party
+import frictionless
+
+# Typing
+from typing import Any, Iterator
+
+
+class MutuallyInclusive(frictionless.Check):
+    """Checks whether mutually inclusive columns are provided together."""
+
+    # Check Attributes
+    code = "mutually-inclusive"
+    Errors = [frictionless.errors.RowConstraintError]
+
+    def __init__(
+        self,
+        descriptor: Any = None,
+        *,
+        field_names: list[str],
+    ) -> None:
+        """Instantiate the MutuallyInclusive Checker"""
+        # Initialise Super Class
+        super().__init__(descriptor)
+
+        # Instance Attributes
+        self.__field_names = field_names
+
+    def validate_row(self, row: frictionless.Row) -> Iterator[frictionless.Error]:
+        """Called to validate the given row (on every row).
+
+        Args:
+            row (frictionless.Row): The row to check the mutual inclusion of.
+
+        Yields:
+            frictionless.Error: For when the mutual inclusion is violated.
+        """
+        # Retrieve Field Names for Missing Cells
+        missing = [f for f in self.__field_names if not row[f]]
+
+        # Check Mutual Inclusivity
+        # Either none of the cells must be missing, or all of them must be
+        # missing, which is checked below by comparing the number of missing
+        # cells against the expected number of cells
+        if len(missing) in (0, len(self.__field_names)):
+            # Short-circuit
+            return
+
+        # Yield Error
+        yield frictionless.errors.RowConstraintError.from_row(
+            row=row,
+            note=(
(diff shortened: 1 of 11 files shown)


tests/plugins/test_mutual_inclusion.py
```diff
@@ -0,0 +1,40 @@
...
+        checks=[
+            plugins.mutual_inclusion.MutuallyInclusive(
+                field_names=["a", "b"],
+            ),
+        ]
+    )
+
+    # Check
+    assert not report.valid
+    assert len(report.flatten()) == 4
```
serge-gaia, line 40: A general remark for all the `Check` tests: it would be better if we could assert which row  is invalid instead of just the count of invalid rows. I don't know if the report allows you to have access at the row number.
Not important for now.

abis_mapping/plugins/mutual_inclusion.py
```diff
@@ -0,0 +1,58 @@
...
+            frictionless.Error: For when the mutual inclusion is violated.
+        """
+        # Retrieve Field Names for Missing Cells
+        missing = [f for f in self.__field_names if not row[f]]
+
+        # Check Mutual Inclusivity
+        # Either none of the cells must be missing, or all of them must be
+        # missing, which is checked below by comparing the number of missing
+        # cells against the expected number of cells
+        if len(missing) in (0, len(self.__field_names)):
```
serge-gaia, line 47: :+1:

abis_mapping/plugins/mutual_inclusion.py
```diff
@@ -0,0 +1,58 @@
...
+            # Short-circuit
+            return
+
+        # Yield Error
+        yield frictionless.errors.RowConstraintError.from_row(
+            row=row,
+            note=(
+                f"the columns {self.__field_names} are mutually inclusive and must be provided together "
+                f"(columns {missing} are missing)"
```
serge-gaia, line 56: arguably I would display `values for the columns {missing} are missing`. 
Not important. Just me.

serge-gaia (approved):

serge-gaia:
Question: does the frictionless framework provides a way to declare the 'Checks` in the schema.  
For example can we put somewhere in the `schema.json` that we want `mutually-inclusive('threatStatus`, `conservationJurisdiction')`?

abis_mapping/plugins/tabular.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+# Third-Party
+import frictionless
+
+# Typing
+from typing import Iterator
+
+
+class NotTabular(frictionless.Check):
```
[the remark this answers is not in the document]
HaydenGaia: That's a good point, not sure why I named it `NotTabular`. Will fix that :+1:

HaydenGaia:
> Question: does the frictionless framework provides a way to declare the checks in the schema. For example can we put somewhere in the `schema.json` that we want `mutually-inclusive('threatStatus`, `conservationJurisdiction')`?

I would prefer it that way as well. I spent a bit of time looking into how I could do it that way, but it doesn't seem like something that frictionless supports at the moment. Even adding extra "constraints" to the `schema.json` isn't really supported by frictionless.

Of course, in future if we find a way to do it then I am happy to change it :+1:

abis_mapping/plugins/tabular.py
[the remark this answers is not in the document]
HaydenGaia: Fixed in: https://github.com/gaiaresources/abis-mapping/pull/41

PR #42, opened 2022-09-14
BDRC-66: Vocab Improvements

HaydenGaia:
## Summary
* Adds alternate labels to new vocabularies
    * `threatStatus` (`vocabs.threat_status`)
    * `threatStatusCheckProtocol` (`vocabs.check_protocol`)
    * `conservationJurisdiction` (`vocabs.conservation_jurisdiction`)
* Adds functionality to match a wider range of labels to vocabulary terms using label normalisation / sanitation.
    * For example, `basisOfRecord/HumanObservation` will now match with the input values `{"Human Observation", "human-observation", "HUMAN_OBSERVATION", etc...}` in the template _without_ explicitly defining those as alternate labels

## Changes
* Adds the `utils.strings.sanitise()` utility function
    * Capitalises and strips all non-alphanumeric characters from a string
    * This is used to normalize the vocabulary term labels
    * This means vocabulary terms can be matched more broadly by only matching on normalised/sanitised values
* Removes overly-verbose "duplicate" alternate labels from all vocabulary terms
    * Similar labels (e.g., `-` instead of `_`, etc) are now handled by the string sanitation instead of being repeatedly and explicitly defined
* Adds new unit tests

### DIFF
--- a/abis_mapping/utils/strings.py
+++ b/abis_mapping/utils/strings.py
@@ -1,0 +1,25 @@
+"""Provides string utilities for the package"""
+
+
+# Standard
+import re
+
+
+# Constants
+REGEX_STRIP = re.compile(
+    pattern=r"[\W_]+",
+    flags=re.IGNORECASE | re.UNICODE,
+)
+
+
+def sanitise(value: str) -> str:
+    """Capitalises and strips all non-alphanumeric characters from a string.
+
+    Args:
+        value (str): String to be stripped and capitalised
+
+    Returns:
+        str: The stripped and capitalised string
+    """
+    # Strip, Capitalise and Return
+    return REGEX_STRIP.sub("", value.upper())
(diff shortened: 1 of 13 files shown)


HaydenGaia:
@serge-gaia Ready for review

abis_mapping/utils/strings.py
```diff
@@ -0,0 +1,25 @@
...
+    """Capitalises and strips all non-alphanumeric characters from a string.
+
+    Args:
+        value (str): String to be stripped and capitalised
+
+    Returns:
+        str: The stripped and capitalised string
+    """
+    # Strip, Capitalise and Return
+    return REGEX_STRIP.sub("", value).upper()
```
serge-gaia, line 25: Haha, that got me into thinking about collision of meaning: English words that once concatenated mean something different. Example: together -> to get her :smile: 
(from https://www.reddit.com/r/wordplay/comments/ghqpeh/words_that_can_be_split_into_two_words/)

I don't think it is a problem here :smile:

serge-gaia:
:+1: on simplification

PR #47, opened 2022-11-17
BDRNS-56: bypass frictionless memory check for Resource.validate() by setting memory limit = 0.

serge-gaia:


HaydenGaia (approved):
Looks good, thanks for making this change :smile:

PR #48, opened 2022-12-01
BDRNS-68: Remove Australia boundaries check in validation but restrict a latitude and longitude min/max values

serge-gaia:
* Removed the ValidCoordinates check (check Australia mainland boundaries).
* schema.json: set latitude maximum = 0 instead -90 and longitude minimum = 0 instead of -180 in order to catch sign mistakes.
* bumped version to 2.2.0


HaydenGaia (approved):
:+1:

PR #53, opened 2023-09-11
BDRSPS-129 created chronological order custom frictionless check

joecrowleygaia:
This is a required check to occur with the upcoming survey metadata template.


abis_mapping/plugins/chronological.py
```diff
@@ -0,0 +1,61 @@
...
+                dts.append(val)
+
+        # Validate chronological order of the list
+        chrono_valid = timestamps.is_chronologically_ordered(dts)
+
+        # Check validity
+        if not chrono_valid:
+            yield frictionless.errors.RowConstraintError.from_row(
+                row=row,
+                note="the provided dates are not in correct chronological order"
```
serge-gaia, line 60: I will be good to display the field_names and also maybe the values in the error note.  
Something like:
```
f"the following dates are not in chronological order: {self.__field_names}. Values are: {dts}"
```

abis_mapping/plugins/chronological.py
```diff
@@ -0,0 +1,61 @@
...
+            frictionless.Error: When the chronological order is violated.
+        """
+        # Get dates or datetimes
+        dts: list[types.DateOrDatetime] = []
+
+        # Populate dates list
+        for name in self.__field_names:
+            val = row[name]
+            if val is not None:
+                dts.append(val)
```
serge-gaia, line 51: This could shorten into a one-line list comprehension but I'm fine as it is.

tests/plugins/test_chronological.py
```diff
@@ -0,0 +1,40 @@
...
+            # Missing
+            {"name": "A", "start": "2022-09-11T15:15:15Z", "end": None},
+            {"name": "B", "start": None, "end": "2022-09-11T15:15:15Z"},
+            {"name": "C", "start": None, "end": None},
+
+            # Valid
+            {"name": "D", "start": "2022-09-11T15:15:15Z", "end": "2023-09-11T15:15:15Z"},
+
+            # Invalid
+            {"name": "E", "start": "2023-09-11T15:15:15Z", "end": "2022-09-11T15:15:15Z"},
```
serge-gaia, line 25: This is a special case of datetime being equal. Maybe name the test accordingly and add the more common case of date.
```
{"name": "End date before start date is invalid, "start": "2023-09-11", "end": "2022-09-01"},
{"name": "Same Date/Time is invalid", "start": "2023-09-11T15:15:15Z", "end": "2022-09-11T15:15:15Z"}
```

tests/plugins/test_chronological.py
```diff
@@ -0,0 +1,40 @@
...
+        checks=[
+            plugins.chronological.ChronologicalOrder(
+                field_names=["start", "end"]
+            ),
+        ]
+    )
+
+    # Assert
+    assert not report.valid
+    assert len(report.flatten()) == 1
```
serge-gaia, line 40: Maybe we could do a bit more by checking the row number, just to be sure that the failures are where we expect them :smile:

serge-gaia (approved):
All good but two very minor points (see comments):
* Add the field names and values in the validation error note.
* Add a couple of valid and invalid test cases involving just dates (no time)

abis_mapping/plugins/chronological.py
joecrowleygaia: updated to use the selected wording

abis_mapping/plugins/chronological.py
joecrowleygaia: updated to use list comprehension

tests/plugins/test_chronological.py
```diff
@@ -0,0 +1,40 @@
...
+    # Construct fake resource
+    resource = frictionless.Resource(
+        source=[
+            # Missing
+            {"name": "A", "start": "2022-09-11T15:15:15Z", "end": None},
+            {"name": "B", "start": None, "end": "2022-09-11T15:15:15Z"},
+            {"name": "C", "start": None, "end": None},
+
+            # Valid
+            {"name": "D", "start": "2022-09-11T15:15:15Z", "end": "2023-09-11T15:15:15Z"},
```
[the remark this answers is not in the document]
joecrowleygaia: added 2 more test case tables and now doing checks against the number of errors expected.

tests/plugins/test_chronological.py
joecrowleygaia: same dates are valid. added same date situation to 3rd test case table.

tests/plugins/test_chronological.py
joecrowleygaia: added multiple dates to the list of scenarios to the test function.

serge-gaia (approved):

PR #55, opened 2023-09-14
BDRSPS-129: Fix mypy errors introduced

joecrowleygaia:
This will fix the mypy errors inadvertently introduced in testing files surrounding the chronological check mainly.


serge-gaia:
i should have check that

serge-gaia (approved):

PR #56, opened 2023-09-14
BDRSS-270: Upgraded RDFLib to V7 #Needs update to BDRPartA

joecrowleygaia:
This will update rdflib from v6.1.1 to v7.0.0. However there should be an update of rdflib for BDRpartA before it can be merged otherwise it will cause conflict between the two.


tests/utils/test_rdf.py
```diff
@@ -16,9 +16,14 @@ def test_rdf_create_graph() -> None:
...
-    # Check Graph
+    # Check Graph returned
     assert isinstance(graph, rdflib.Graph)
-    assert len(list(graph.namespaces())) == 30
+
+    # Check namespaces correctly assigned
+    namespaces = list(graph.namespaces())
+    required_namespaces = [(name, rdflib.term.URIRef(str(uri))) for (name, uri) in utils.rdf.REQUIRED_NAMESPACES]
+    for ns in required_namespaces:
+        assert ns in namespaces
```
serge-gaia, line 26: :+1:

serge-gaia (approved):

PR #58, opened 2023-10-13
BDRSPS-190 Update Frictionless to V5

joecrowleygaia:
A lot of changes were necessary to get to this point, of note:

- TableResource now used in place of Resource
- Refactoring of custom field types since the Type class was removed
- Row numbering changed to throw off our code that relied upon it e.g. chunking, URI setting
- "list" type field delimiter now specified using "delimiter" property (default still pipe "|"), no longer relying on regex of field type name previously deployed e.g."list[|]"

Refer to YouTrack [BDRSPS-190](https://youtrack.gaiaresources.com.au/youtrack/issue/BDRSPS-190/Backend-update-to-frictionless-V5) for more details


abis_mapping/templates/survey_metadata/mapping.py
```diff
@@ -40,28 +41,31 @@ def apply_validation(
         Returns:
             frictionless.Report: Validation report for the specified data.
         """
-        # Construct Resource (Table with Schema)
+        # Construct Schema
+        schema = frictionless.Schema.from_descriptor(self.schema())
+
+        # Construct TableResource
         resource = frictionless.Resource(
```
serge-gaia, line 48: Silly question: why this is not a `TableResource` here like the others?

joecrowleygaia: No not silly at all. This still works so i think maybe it was another change that may have fixed a previous issue i was having (i think it had something either to do with the keyword arg being `data=data` or using the resource open method). I will see if i can change the use of `TableResource` back to `Resource` and push. Leave it with me.

joecrowleygaia: OK i changed the use of `TableResource` back to plain old `Resource` which still works as intended. The comments within the `apply_mapping` and `apply_validation` methods suggests that a more generic resource is intended and use of `Resource` is more appropriate allowing for easier expansion beyond tabular data.

serge-gaia (commented):
Minor but could you review the lint and typecheck warnings.
```
+ flake8 tests abis_mapping
tests/plugins/test_list.py:135:1: W391 blank line at end of file
tests/utils/test_chunking.py:5:1: F401 'frictionless' imported but unused
(abis-mapping-py3.9) serge@serge-Dell:~/Projects/abis-mapping$ ./scripts/typecheck.sh 
+ mypy tests abis_mapping
tests/plugins/test_timestamp.py:36: error: Self? has no attribute "get_field"  [attr-defined]
tests/plugins/test_list.py:23: error: Self? has no attribute "get_field"  [attr-defined]
Found 2 errors in 2 files (checked 61 source files)
```
For the typecheck `no attribute "get_field"`, I think you just need to type the var `schema: frictionless.Schema`

joecrowleygaia:
fixed flake8 and mypy issues as noted.

> Minor but could you review the lint and typecheck warnings.
> 
> ```
> + flake8 tests abis_mapping
> tests/plugins/test_list.py:135:1: W391 blank line at end of file
> tests/utils/test_chunking.py:5:1: F401 'frictionless' imported but unused
> (abis-mapping-py3.9) serge@serge-Dell:~/Projects/abis-mapping$ ./scripts/typecheck.sh 
> + mypy tests abis_mapping
> tests/plugins/test_timestamp.py:36: error: Self? has no attribute "get_field"  [attr-defined]
> tests/plugins/test_list.py:23: error: Self? has no attribute "get_field"  [attr-defined]
> Found 2 errors in 2 files (checked 61 source files)
> ```
> 
> For the typecheck `no attribute "get_field"`, I think you just need to type the var `schema: frictionless.Schema`
> 
> Apart that, all good.

serge-gaia:
Thanks

serge-gaia (approved):

PR #59, opened 2023-10-17
BDRSPS-174 Allow extra columns in CSV Validation

joecrowleygaia:
This relies on pull request [58](https://github.com/gaiaresources/abis-mapping/pull/58) to be merged as the changes rely on frictionless v5


tests/base/test_mapper.py
```diff
@@ -91,3 +93,32 @@ def test_base_get_instructions() -> None:
...
+    assert mapper is not None
+
+    # Ingest extra column data
+    data = pathlib.Path(file_path).read_bytes()
+
+    # Perform validation
+    report = mapper().apply_validation(data)
+
+    # Assert
+    assert report.valid
```
serge-gaia, line 124: :+1: 

A suggestion. We could add a test that we expect to fail showing that if one or more extra columns are not placed at the end, the validation fails. Just to show we are aware of the issue.

joecrowleygaia: Have added tests as requested

serge-gaia: Thank you.

serge-gaia (approved):

PR #60, opened 2023-10-20
BDRSPS-175: Extra fields csv parsing

joecrowleygaia:
Created two extra methods to achieve this

- One that takes either raw data or a frictionless row and produces a "diff" schema
- Another that takes a frictionless row, produces diff schema from it then returns a dictionary of the extra values

### DIFF
--- a/abis_mapping/base/mapper.py
+++ b/abis_mapping/base/mapper.py
@@ -89,13 +89,14 @@ class ABISMapper(abc.ABC):
     @functools.lru_cache
     def extra_columns_schema(
         cls,
-        data: types.ReadableType,
+        data: types.ReadableType | frictionless.Row,
     ) -> frictionless.Schema:
         """Creates and caches a schema with all extra fields found in data.
 
         Args:
-            data (types.ReadableType): Readable raw csv data, expected to contain
-                more columns than included in the template's schema.
+            data (types.ReadableType | frictionless.Row): Readable raw csv
+                data or frictionless row, expected to contain more columns
+                than included in the template's schema.
 
         Returns:
             frictionless.Schema: A schema object, the fields of which are only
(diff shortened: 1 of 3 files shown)


abis_mapping/base/mapper.py
```diff
@@ -66,6 +66,76 @@ def apply_mapping(
...
+        extra_schema = cls.extra_fields_schema(row)
+
+        # Create dictionary consisting row data from extra fields only
+        return {field: row[field] for field in extra_schema.field_names if row[field] is not None}
+
+    @final
+    @classmethod
+    def extra_fields_schema(
+        cls,
+        data: types.ReadableType | frictionless.Row
```
serge-gaia, line 93: Ha! This new union type notation (`|` instead `Union[..]` )  is python 3.10 and above only. We are still stuck with python 3.9 atm with BDR-partA.  
I would recommend to forgo `types.ReadableType` as an acceptable type for now, see comment below.

serge-gaia (changes requested):

abis_mapping/base/mapper.py
```diff
@@ -66,6 +66,76 @@ def apply_mapping(
...
+        if isinstance(data, frictionless.Row):
+            # Get set of fieldnames of row
+            actual_fieldnames = set(data.field_names)
+
+            # Create schema from row fields
+            actual_schema = frictionless.Schema(fields=data.fields)
+
+        else:
+            # Construct resource from data and infer
+            resource = frictionless.Resource(
```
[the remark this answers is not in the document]
joecrowleygaia: I originally had it planned in my mind to find the extra columns before iterating over rows, but then saw that would come with potentially more complexity when it came time to implement it in the apply_mapping method. I then implemented the row-wise technique. I'm not sure which method will be more optimised but if the mapping becomes the bottleneck this could be something to target potentially. 

I have removed the readableType method overloading.

abis_mapping/base/mapper.py
joecrowleygaia: Removed the ReadableType method overloading. See comment above or below about potential future optimisation.

serge-gaia (approved):

PR #63, opened 2023-10-26
BDRSPS-177 Extra fields JSON Graph Building

joecrowleygaia:
It now adds extra fields as a json string literal to the graph for each of the three existing templates.
Some changes to note:

- Had to go back to accepting raw data into the extra schema method, as it is required in setting up the resource before iterating over its rows.
- The extra fields are now found by slicing the field_names list rather than using the difference of sets approach, to preserve field ordering
-  The extra schema method now either returns the full schema or just the difference, controlled via the full_schema flag (default: False).

### DIFF
--- a/tests/templates/test_survey_metadata.py
+++ b/tests/templates/test_survey_metadata.py
@@ -1,5 +1,6 @@
 # Third-party
 import frictionless
+import pytest
 
 # Standard
 import pathlib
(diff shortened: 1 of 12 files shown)


tests/templates/test_survey_metadata.py
```diff
@@ -71,6 +81,14 @@ def test_mapping() -> None:
         graph2=expected,
     )
 
+    # Check that there are no `None`s in the Graph
+    # This check is important. As some fields are optional they can be `None`
+    # at runtime. Unfortunately, `None` is valid in many contexts in Python,
+    # including string formatting. This means that type-checking is unable to
+    # determine whether a statement is valid in our specific context. As such,
+    # we check here to see if any `None`s have snuck their way into the RDF.
+    assert "None" not in graphs[0].serialize(format="ttl")
```
serge-gaia, line 90: :+1:

serge-gaia (approved):
Nice one.

PR #66, opened 2023-10-31
BDRSPS-238 Created Custom WKT Field Type

joecrowleygaia:
This creates custom WKT Field type which relies on Shapely for parsing and validation. Since there are no type annotations included with shapely and no stubs available through typeshed, the choice was made to suppress errors this introduces by adding an exception to the mypy config, similar to what was adopted for frictionless previously.


abis_mapping/plugins/wkt.py
```diff
@@ -0,0 +1,105 @@
...
+
+            Returns:
+                str: Converted cell.
+
+            Raises:
+                TypeError: if the returned value from the shapely
+                    method to_wkt is not string.
+            """
+            # Serialize to default format for pyproj
+            wkt_str = shapely.to_wkt(cell)
```
serge-gaia, line 92: Shouldn't we catch exception here or guard against argument not being a `Geometry`?
It's probably never going to happen, the way we use it though.

tests/plugins/test_wkt.py
```diff
@@ -0,0 +1,63 @@
...
+    assert field.read_cell("POLYGON ((1 1, 1 4, 4 4, 4 1))")[0] is None
+
+    # Read valid cells
+    assert field.read_cell("POINT (2.0 3.0)")[0]
+    assert field.read_cell("LINESTRING (1 1, 2 2, 3 3)")[0]
+    assert field.read_cell("POLYGON ((1 1, 1 4, 4 4, 4 1, 1 1))")[0]
+    assert field.read_cell("MULTIPOINT (2 3, 4 5, 6 7)")[0]
+    assert field.read_cell("MULTILINESTRING ((1 1, 2 2, 3 3), (4 4, 5 5))")[0]
+    assert field.read_cell("MULTIPOLYGON (((0 0, 0 1, 1 1, 1 0, 0 0)), ((2 2, 2 3, 3 3, 3 2, 2 2)))")[0]
+    assert field.read_cell("GEOMETRYCOLLECTION (POINT (2 3), LINESTRING (1 1, 2 2))")[0]
```
serge-gaia, line 59: :+1

abis_mapping/plugins/wkt.py
serge-gaia: I'm thinking about the `rounding_precision` argument for this method. It is set to `6` by default which is perfectly fine in my opinion (11 cm) but I know that people love their decimals even if it doesn't mean anything.

```python
shapely.to_wkt(shapely.from_wkt("POINT(0.12345678 0.12345678)"))
'POINT (0.123457 0.123457)'
```
by putting `rounding_precision = 8` (and `trim=False` (default) we have:

```python
shapely.to_wkt(shapely.from_wkt("POINT(0.12345678 0.12345678)"), rounding_precision=8)
'POINT (0.12345678 0.12345678)'
```
and also we don't change/extend shorter numbers:

```python
shapely.to_wkt(shapely.from_wkt("POINT(0.12345 0.12345)"), rounding_precision=8)
'POINT (0.12345 0.12345)'
```

I think we should put `rounding_precision = 8`  so we should not have complaints about us shortening the coordinates.

What do you think @joecrowleygaia ?

joecrowleygaia: Another alternative would be to set to -1 which is full precision. I'll change to 8 for now but simple enough for me to change again if needed.

joecrowleygaia: i reviewed the `value_writer` methods for the other field types both within the frictionless framework and what we have currently and i didn't find guarding of arguments provided to these methods. The type checking i did do for the `wkt_str` variable has to do with the config for shapely in mypy.ini ignoring types which sets everything to `Any` that comes out of it in the eyes of the type checker.

serge-gaia: ok

joecrowleygaia: pushed changes for your review @serge-gaia

serge-gaia (approved):

PR #73, opened 2023-11-08
Bumped the poetry package version to 3.0.0-a1 (alpha1)

serge-gaia:
Bumping the version to 3.0.0-a1 (alpha1) so that we can visually see if the deployment of partA is using the right version (in the swagger)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "abis-mapping"
-version = "2.2.0"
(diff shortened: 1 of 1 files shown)


joecrowleygaia (approved):
does this mean a tag and release needs to be generated also?

serge-gaia:
not at this stage. Probably in sprint 6 or 7 when we do release on UAT. This just a visual cue for me to see if the CI/CD deployment works on the chunkinator environment.

PR #74, opened 2023-11-13
BDRSPS-284 Created logical OR check

joecrowleygaia:
This will provide the necessary schema check required to implement the SSD site template providing a mechanism to check for either `footprintWKT` and `decimalLat/Long` are present for validation.


joecrowleygaia:
Refactored the code a little removing the exclusive OR behaviour from the check as it wasn't required.

serge-gaia (approved):

PR #75, opened 2023-11-14
BDRSPS-239 Added Systematic Survey Site Schema

joecrowleygaia:
This adds the survey site schema as well as the validation associated with it. Some refactoring of mapper tests occurred to happen via `mark.parametize` rather than looping within test functions, given the list of templates now becoming larger, potentially leading to scenario of multiple templates failing but being hidden by the raising of assertion error only on first encountered problem.


abis_mapping/templates/survey_site_data/schema.json
```diff
@@ -0,0 +1,142 @@
...
+    },
+    {
+      "name": "siteVisitStart",
+      "title": "Site Visit Start",
+      "description": "The temporal start of when the Site was being used to collect data for the survey. Expected values include date, dateTime, dateTimeStamp.",
+      "example": "2016-02-28",
+      "type": "timestamp",
+      "format": "default",
+      "constraints": {
+        "required": true
```
serge-gaia, line 126: I have a doubt of this being required but it is not the place to discuss about it.

serge-gaia (approved):

abis_mapping/templates/survey_site_data/metadata.json
```diff
@@ -0,0 +1,13 @@
+{
+  "id": "survey_site_data.csv",
+  "label": "Systematic Survey Site Data Template",
+  "version": "0.1.0",
+  "description": "A template for systematic survey site data",
+  "biodiversity_type": "Biological Systematic Survey Site Data",
+  "spatial_type": "N/A",
```
serge-gaia, line 7: Minor but here we could put: `Lat/Long or WKT`

serge-gaia: or 'Any'

serge-gaia: or `Point, Polygon`

PR #79, opened 2023-11-23
BDRSPS-272 Added site id as a rdfs comment for survey occurrences template mapping

joecrowleygaia:


abis_mapping/templates/survey_occurrence_data/mapping.py
```diff
@@ -1116,6 +1116,11 @@ def add_sampling_field(
             # Add Basis Of Record Attribute
             graph.add((uri, utils.namespaces.TERN.hasAttribute, basis))
 
+        # Check for siteID
+        if site_id := row["siteID"]:
+            # Add site id comment
+            graph.add((uri, rdflib.RDFS.comment, rdflib.Literal(site_id)))
```
serge-gaia, line 1122: I think that Mieke was right. The RDFS.comment should have been a JSON string like:
```
{
  "siteID": "P1"
 }
 ```
 like we do for extra fields.  
 Currently there's no 'visual' indication that is a siteID.  
 Anyhow merging that for now and let continue our discussion.

serge-gaia (approved):

PR #90, opened 2023-12-08
BDRSPS-335 Made survey metadata template valid for one and only one row

joecrowleygaia:
The previous `NotEmpty()` check was removed and replaced with a `min_rows=1` along with `max_rows=1` arguments to the frictionless `table_dimensions()` check.

### DIFF
--- a/abis_mapping/templates/survey_metadata/mapping.py
+++ b/abis_mapping/templates/survey_metadata/mapping.py
@@ -8,6 +8,7 @@ from abis_mapping import utils
 
 # Third-party
 import frictionless
+import frictionless.checks
 import rdflib
 
 # Typing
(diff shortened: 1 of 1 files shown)


abis_mapping/templates/survey_metadata/mapping.py
```diff
@@ -52,9 +53,11 @@ def apply_validation(
         report: frictionless.Report = resource.validate(
             checklist=frictionless.Checklist(
                 checks=[
+                    # Enforces non-empty and maximum row count.
+                    frictionless.checks.table_dimensions(max_rows=1, min_rows=1),
```
serge-gaia, line 57: :+1:

serge-gaia (approved):

PR #91, opened 2023-12-14
BDRSPS-350 Made extra fields all strings

joecrowleygaia:
This change makes the additional fields extracted from a csv template all be viewed as a string.


serge-gaia (approved):
Looks good. I don't think I need to test it.

PR #92, opened 2023-12-15
BDRSPS-338 Add year and yearmonth capabilities to the metadata temporalCoverage fields

joecrowleygaia:
This adds the ability to accept and map year and year month dates that are provided in either the `temporalCoverageStartDate` and `temporalCoverageEndDate` survey metadata fields. To achieve this the following occurred:
- Timestamp field type was modified to contain two new attributes `allow_year: bool` and `allow_year_month: bool` which are also settable via the json schema, with a default of `False` to not interfere with other existing timestamp fields.
- The python representation of a yearmonth in this field is the same as a `yearmonth` type field in frictionless i.e. a named tuple with attributes `year` and `month`
- The representation of a year is an `int`
- Modification to comparison of timestamps for the chronological ordering checks was performed to allow offset-naive and offset-aware dates and datetimes to be compared. The assumption used was that a naive datetime has the same timezone as the aware, and so timezone information is removed prior to comparison.
- If a year is compared to any date/datetime with the same year it is deemed to be equal/valid
- If a yearmonth is compared to any date/datetime with the same year and month it is deemed to be equal/valid
- Year month input can be `mm/yyyy` `yyyy-mm` or `m/yyyy` meaning the slash form is zero padding optional
- Year month output is `yyyy-mm`
- Year input and output is `yyyy`
- Acceptable year range matches the same used within frictionless `0 <= year <= 9999`

Further improvement i think could be made by converting the Timestamp type to a class, bringing all the disparate methods in together as well allowing use of magic methods e.g. `__repr__` , `__lt__` and `__gt__` which should help keep the code more readable in the future.


serge-gaia (approved):
Good job with the tests.

PR #113, opened 2024-02-14
BDRSPS-435 Template IDs now include version number.

joecrowleygaia:
revised template id to now be generated from metadata json. Version number now included in template id.
This is in preparation to allow for running multiple versions of templates at the same time.


chungvl (approved):

chungvl (approved):
Looks OK - be aware partA/backend will need work to fix unit tests.

PR #114, opened 2024-02-14
Revert "BDRSPS-435 Template IDs now include version number."

chungvl:
Reverts gaiaresources/abis-mapping#113


joecrowleygaia (commented):
I believe unnecessary. Staging now appears to be working correctly.

PR #120, opened 2024-03-12
BDRSPS-494 New site template geometry validation

joecrowleygaia:
This implements the agreed validation rules for the site template, incorporating whether the geodetic datum is included or not. A new custom frictionless check was created for this purpose specifically, hence the hard-coding of field names within the check's `validate_row` method.

### DIFF
--- a/abis_mapping/plugins/sites_geometry.py
+++ b/abis_mapping/plugins/sites_geometry.py
@@ -46,7 +46,7 @@ class SitesGeometry(frictionless.Check):
             return
 
         # Create error note
-        note = f"invalid geometry: "
+        note = "invalid geometry: "
         if (wkt or (lat and long)) and not datum:
             note += f"invalid datum {row['geodeticDatum']} provided."
         elif lat ^ long:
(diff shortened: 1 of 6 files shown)


abis_mapping/plugins/sites_geometry.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        # Extract values as bool
+        lat = row["decimalLatitude"] is not None
+        long = row["decimalLongitude"] is not None
+        datum = row["geodeticDatum"] is not None
+        wkt = row["footprintWKT"] is not None
+        site_id = row["siteID"] in self.occurrence_site_ids
+
+        # Perform check
+        if (
```
chungvl, line 41: Remove outer parentheses?

chungvl (changes requested):
Tiny edit requested

abis_mapping/plugins/sites_geometry.py
joecrowleygaia: i find it easier to read as a multi-line rather than a single-line. I put each AND group on a new line here. Lemme know if you would prefer the other way and I will remove the parentheses.

chungvl (approved):

PR #127, opened 2024-03-27
BDRSPS-534 Create shape graph for sites' template

joecrowleygaia:
I created this using rdflib. There is no functional usage of the SHACL in pytest testing suites yet or any other reason yet but provides a base file to be updated worked on for future work


serge-gaia (approved):
Looks good. When I run it my `validation.ttl` is different but it's just orders of the triples. RDFLib graph serialisation order is kind of random.

PR #132, opened 2024-04-10
BDRSPS-463: Creation of Survey Site Data v1.0

chungvl:


joecrowleygaia (changes requested):
I don't think we are maintaining multiversions of the SSD templates yet. Please put all changes in the existing `survey_site_data/` dir for now. I suspect next sprint or the one after we will start heading down this road. Version should be whatever it is currently (v0.x.x).

serge-gaia:
Yes, only one version for SSD. Two versions for incidental only.

abis_mapping/templates/survey_site_data/metadata.json
```diff
@@ -7,7 +7,7 @@
   "spatial_type": "Point, line, polygon",
   "file_type": "CSV",
   "sampling_type": "systematic survey",
-  "template_url": "https://raw.githubusercontent.com/gaiaresources/abis-mapping/main/abis_mapping/templates/survey_site_data/survey_site_data.csv",
-  "schema_url": "https://raw.githubusercontent.com/gaiaresources/abis-mapping/main/abis_mapping/templates/survey_site_data/schema.json",
-  "instructions_url": "https://raw.githubusercontent.com/gaiaresources/abis-mapping/main/abis_mapping/templates/survey_site_data/instructions.pdf"
+  "template_url": "https://raw.githubusercontent.com/gaiaresources/abis-mapping/main/abis_mapping/templates/survey_site_data_v1/survey_site_data.csv",
```
joecrowleygaia, line 10: update path

abis_mapping/templates/survey_site_data/metadata.json
```diff
@@ -7,7 +7,7 @@
...
+  "schema_url": "https://raw.githubusercontent.com/gaiaresources/abis-mapping/main/abis_mapping/templates/survey_site_data_v1/schema.json",
```
joecrowleygaia, line 11: update path

abis_mapping/templates/survey_site_data/metadata.json
```diff
@@ -7,7 +7,7 @@
...
+  "instructions_url": "https://raw.githubusercontent.com/gaiaresources/abis-mapping/main/abis_mapping/templates/survey_site_data_v1/instructions.pdf"
```
joecrowleygaia, line 12: update path

abis_mapping/templates/survey_site_data/schema.json
```diff
@@ -45,6 +56,212 @@
...
+          "TIDAL FLAT",
+          "VOLCANO"
+      ]
+      }
+    },
+    {
+      "name": "landformElement",
+      "title": "Landform Element",
+      "description": "A description of landform at a finer scale (best mapped at 1:15,000) based on slope, morphological type, dimensions, mode of geomorphological activity and geomorphological agent.",
+      "type": "string",
```
joecrowleygaia, line 116: should be `list`

abis_mapping/templates/survey_site_data/schema.json
```diff
@@ -45,6 +56,212 @@
...
+      }
+    },
+    {
+      "name": "aspect",
+      "title": "Aspect",
+      "description": "The aspect of the location where the observations was made measured in degrees clockwise from true North.",
+      "type": "integer",
+      "example": "30",
+      "format": "default"
+
```
joecrowleygaia, line 219: should be optional

tests/templates/conftest.py
```diff
@@ -106,7 +106,7 @@ class TemplateTestParameters:
         ],
     ),
     TemplateTestParameters(
-        template_id="survey_site_data-v0.1.0.csv",
+        template_id="survey_site_data-v1.0.0.csv",
```
joecrowleygaia, line 109: doesn't match the template version number and tests fail.

abis_mapping/templates/survey_site_data/schema.json
chungvl: I'm getting: 

`frictionless.exception.FrictionlessException: [schema-error] Schema is not valid: descriptor is not valid (Field is not valid: constraint "enum" is not supported by type "list")`

serge-gaia: interesting test case

joecrowleygaia: the `list` field plugin is likely going to need to be updated to handle enums. It can be tied into the string field type's handling of enums hopefully.

joecrowleygaia: ok i think what might have to happen is append "enum" to the `ListField.supported_constraints` list attribute. Then overload the definition of [`check_enum`](https://github.com/frictionlessdata/frictionless-py/blob/75be5990da82965e7c02f3d5309c1579e34f84c3/frictionless/schema/field.py#L338) to perform checks on a per element basis.

joecrowleygaia: as discussed please set to list and remove enum constraint. At the same time please capture this in a backlog ticket to be rectified.

joecrowleygaia (changes requested):

joecrowleygaia (approved):

PR #140, opened 2024-04-18
BDRSPS-572 Changed site template field "Condition" to non capitalised.

joecrowleygaia:


joecrowleygaia:
I will need to update partA backend after merge.

chungvl (approved):

PR #146, opened 2024-04-26
BDRSPS-589: renamed fields visitOrganisations and surveyOrganisation to visitOrgs and surveyOrgs respectively.

joecrowleygaia:
Another thing to consider is that the fields are pluralised as they are list types. PartA backend will need to be updated following this merge.


serge-gaia:
Could you also update the CHANGELO.md
https://github.com/gaiaresources/abis-mapping/blob/main/CHANGELOG.md

joecrowleygaia:
> Could you also update the CHANGELO.md https://github.com/gaiaresources/abis-mapping/blob/main/CHANGELOG.md

I've updated it, but wasn't sure if it should be included as it is for SSD templates.

serge-gaia:
My bad, these changes didn't apply to the incidental occurrences. There were no needs to update the changelog which was just for incidental occurrences so far but what you did is still good.

joecrowleygaia:
> My bad, these changes didn't apply to the incidental occurrences. There were no needs to update the changelog which was just for incidental occurrences so far but what you did is still good.

Reverted the changelog commit. Should be good to go.

serge-gaia:
Sorry about that @joecrowleygaia

serge-gaia (approved):

PR #160, opened 2024-06-18
BDRSPS-697 Added mkdocs-material and github actions for publishing instructions.

joecrowleygaia:
This will hopefully trigger a github actions workflow and the docs should be viewable at https://gaiaresources.github.io/abis-mapping/


.gitignore
```diff
@@ -81,3 +81,6 @@ Dockerfile
 # Generated documents for testing
 /result.csv
 /result.md
+/docs/pages/**/*.md
```
serge-gaia, line 84: Do we exclude these because the `instructions.md`s come from elsewhere (`/abis_mapping/templates`) ?
If yes , shouldn't we be more specific and put instructions.md in the exclude rule. We might want to keep other `.md` than `index.md` in the `/docs` folder.

joecrowleygaia: i wanted to exclude it because the instruction page can be named anything at the moment. In the github action it will create a file called incidental_occurrence_data_v2.md.
I thought for now we could leave it at this and as the other templates get their instructions generated we will have more of an idea how to best structure this better.

serge-gaia: ok

.github/workflows/ci.yml
```diff
@@ -0,0 +1,32 @@
...
+          python-version: 3.11
+      - run: echo "cache_id=$(date --utc '+%V')" >> $GITHUB_ENV
+      - uses: actions/cache@v4
+        with:
+          key: mkdocs-material-${{ env.cache_id }}
+          path: .cache
+          restore-keys: |
+            mkdocs-material-
+      - run: |
+          pipx install poetry
```
serge-gaia, line 28: I think we should set a version of poetry (we are using poetry 1.6 or 1.8) . I don't want the the action to suddenly fail because poetry released a 2.x that breaks our process.

joecrowleygaia: Good call. Done.

serge-gaia (approved):

PR #163, opened 2024-06-20
BDRSPS-701 Fixed pipeline issues

joecrowleygaia:


.github/workflows/ci.yml
```diff
@@ -17,16 +17,18 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: 3.11
```
serge-gaia, line 19: When looking at this job: https://github.com/gaiaresources/abis-mapping/actions/runs/9592649413/job/26451585975

I see 
```
The currently activated Python version 3.10.12 is not supported by the project (^3.11).
```
strange that the currently activated Python version is 3.10.x
It then switch back to 3.11 at the end

joecrowleygaia: i know i don't like it either, and i don't understand the `actions/setup-python@v5` action well enough to know what is happening at this stage.

chungvl (approved):

PR #184, opened 2024-08-20
BDRSPS-736 Fixed bad anchor refs

joecrowleygaia:
Also altered the debug level of all bad links during build to 'warn' and set the build mode to 'strict' for mkdocs/mike.


Lincoln-GR (approved):
Nice :+1:

PR #187, opened 2024-08-21
BDRSPS-712 Adds Darwin core links to field names in markdown tables

joecrowleygaia:
Also cleans up some code duplication for the tabler classes.

### DIFF
--- a/docs/tables/base.py
+++ b/docs/tables/base.py
@@ -4,12 +4,13 @@
 import abc
 import csv
 import io
+import typing
 
 # Local
 from abis_mapping import base
 
 # Typing
-from typing import IO, Final, Any, Mapping, Iterable, Literal, Annotated
+from typing import IO, Final, Any, Mapping, Iterable, Literal
 
 
 class BaseTabler(abc.ABC):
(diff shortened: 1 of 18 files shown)


docs/tables/base.py
```diff
@@ -12,29 +13,66 @@
 
 
 class BaseTabler(abc.ABC):
+    """Base class for tablers.
+
+    Attributes:
+        alignment (list[str] | None): Optional alignment of table columns.
+    """
+    # Class attributes
+    alignment: list[str] | None = None
```
Lincoln-GR, line 22: If this is a class variable (as opposed to instance variable) the type should be wrapped in `ClassVar[...]`

docs/tables/base.py
```diff
@@ -12,29 +13,66 @@
...
+
     def __init__(
         self,
-        template_id: str
+        template_id: str,
+        format: str = "csv",
```
Lincoln-GR, line 27: Should this default to "markdown" as per the docstring? or should the docstring change?

Also you could use the type `Literal["markdown", "csv"]` instead of `str`

Lincoln-GR (commented):
Looks good, I have a couple of small comments

docs/tables/base.py
joecrowleygaia: No requirement for this, should only be utilised if class variables should not be set on an instance. The comment indicates that the attributes available at the class level are declared below, not that they are exclusively set by a class.

docs/tables/base.py
joecrowleygaia: good suggestion. I have added an alias and way to extract the args to perform runtime checks still.

Lincoln-GR (approved):

PR #189, opened 2024-08-23
BDRSPS-772 Defining settings using pydantic_settings model. Created metadata model.

joecrowleygaia:
This will enable the redirection of abis-mapping instructions when a specific versions is used in BDR-partA backend or elsewhere.

### DIFF
--- a/abis_mapping/base/mapper.py
+++ b/abis_mapping/base/mapper.py
@@ -279,10 +279,12 @@ class ABISMapper(abc.ABC):
         directory = pathlib.Path(inspect.getfile(cls)).parent
         metadata_file = directory / "metadata.json"
 
-        # Read Metadata and Return
-        md: dict[str, str] = json.loads(metadata_file.read_text())
-        md["id"] = f"{md['name']}-v{md['version']}.{md['file_type'].lower()}"
-        return md
+        # Read Metadata and validate
+        md_dict = json.loads(metadata_file.read_text())
+        md_class = types.metadata.TemplateMetadata.model_validate(md_dict, strict=True)
+
+        # Return
+        return md_class.model_dump()
 
     @final
     @classmethod
--- a/abis_mapping/plugins/wkt.py
+++ b/abis_mapping/plugins/wkt.py
@@ -93,7 +93,7 @@ class WKTField(frictionless.Field):
             # Serialize to wkt format
             wkt_str = shapely.to_wkt(
                 geometry=cell,
-                rounding_precision=settings.DEFAULT_WKT_ROUNDING_PRECISION,
+                rounding_precision=settings.Settings().DEFAULT_WKT_ROUNDING_PRECISION,
             )
 
             # Type checking due to no types provided by the shapely package
--- a/abis_mapping/settings.py
+++ b/abis_mapping/settings.py
@@ -1,7 +1,22 @@
 """All non-sensitive project-wide configuration parameters"""
 
-# Default precision for rounding WKT coordinates when serializing.
-DEFAULT_WKT_ROUNDING_PRECISION = 8
+# Third-party
+import pydantic_settings
 
-# Default CRS to transform all input geometry to.
-DEFAULT_TARGET_CRS = 'GDA2020'
+
+class Settings(pydantic_settings.BaseSettings):
+    """Class for defining default project-wide settings."""
+    # Default precision for rounding WKT coordinates when serializing.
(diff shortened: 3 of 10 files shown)


abis_mapping/settings.py
```diff
@@ -1,7 +1,22 @@
...
+    DEFAULT_TARGET_CRS: str = 'GDA2020'
+
+    # Base url for the instructions site
+    INSTRUCTIONS_BASE_URL: str = 'https://gaiaresources.github.io/abis-mapping/'
+
+    # The version of the documents to be selected
+    INSTRUCTIONS_VERSION: str = 'dev'
+
+    # If changing via environment variable prefix name with 'ABIS_MAPPING_'
+    model_config = pydantic_settings.SettingsConfigDict(env_prefix="ABIS_MAPPING_")
```
Lincoln-GR, line 22: I think  we don't use env files in prod (right?) - but it might be helpful in dev to have `env_file='.env'` added here?

abis_mapping/plugins/wkt.py
```diff
@@ -93,7 +93,7 @@ def value_writer(cell: shapely.Geometry) -> str:
             # Serialize to wkt format
             wkt_str = shapely.to_wkt(
                 geometry=cell,
-                rounding_precision=settings.DEFAULT_WKT_ROUNDING_PRECISION,
+                rounding_precision=settings.Settings().DEFAULT_WKT_ROUNDING_PRECISION,
```
Lincoln-GR, line 96: Every time `Settings()` is instantiated pydantic has to go and read+parse the env variables, could you add `SETTINGS = Settings()` at the bottom of `settings.py` and then reference that everywhere? (similar to how part a does it)

abis_mapping/base/mapper.py
```diff
@@ -279,10 +279,12 @@ def metadata(cls) -> dict[str, str]:
...
-        # Read Metadata and Return
-        md: dict[str, str] = json.loads(metadata_file.read_text())
-        md["id"] = f"{md['name']}-v{md['version']}.{md['file_type'].lower()}"
-        return md
+        # Read Metadata and validate
+        md_dict = json.loads(metadata_file.read_text())
+        md_class = types.metadata.TemplateMetadata.model_validate(md_dict, strict=True)
+
+        # Return
+        return md_class.model_dump()
```
Lincoln-GR, line 287: Would returning `md_class` here be a good idea to increase the amount of code using the nice `TemplateMetadata` object rather than a plain dict? If its gonna be a big change we can leave it for a future improvement.

poetry.lock
```diff
@@ -1001,141 +1001,254 @@ files = [
...
 annotated-types = ">=0.4.0"
-pydantic-core = "2.10.1"
+pydantic-core = "2.18.4"
 typing-extensions = ">=4.6.1"
 
 [package.extras]
 email = ["email-validator (>=2.0.0)"]
 
+[[package]]
+name = "pydantic"
```
Lincoln-GR, line 1022: pydantic seems to be listed twice - here and line 1003 - do you need to run `poetry lock --no-update` to refresh the lock file?

Lincoln-GR (commented):
Looks good, left some comments

abis_mapping/settings.py
joecrowleygaia: Not intended for env files, rather the setting of environment variables to override these from within the project that utilises the library. BDR PartA will in future utilise this to form the instructions urls on a new release.

abis_mapping/plugins/wkt.py
joecrowleygaia: I had originally had it that way however pydantic_settings documentation has it being processed in this manner. It makes more sense to reprocess the env variables each time for this library.

abis_mapping/base/mapper.py
joecrowleygaia: leave as is

abis_mapping/settings.py
Lincoln-GR: Yes, my suggestion is to add env file lookup so developers have the option of overriding from an env file locally, which is what part A does with its settings.

poetry.lock
joecrowleygaia: ran command and pushed.

Lincoln-GR (approved):

abis_mapping/settings.py
joecrowleygaia: This could be added in a future release. I can't picture us using it at this stage.

PR #197, opened 2024-09-05
Check for warnings in tests and more flake8 rules

Lincoln-GR:
* check that no warnings are emitted during tests
* Enable more flake8 rules to have style closer to black, and enable some optional rules from flake8-bugbear


joecrowleygaia:
I'd prefer to stick with google styling for now unless @serge-gaia tells us to move to black, especially since it has been decided as an action from the software engineering guild and is documented as the style to use for the project in confluence. The formatter i use (apparently google compliant) produces significantly different results.

joecrowleygaia (changes requested):
Wait for @serge-gaia to get back

PR #204, opened 2024-09-12
BDRSPS-802 Revert "added status field to metadata to indicate template stage in its lifecycle."

joecrowleygaia:
This reverts commit 87ecec3a3263f8b5154961112535adba0c6bc733. No longer required.


Lincoln-GR (commented):
@joecrowleygaia Shoud we keep this field around for if we ever want to issue deprecation warnings when submissions are made with deprecated templates?

joecrowleygaia:
i don't believe so, i think it cound add confusion for the time waiting for the implementation of a deprecation warning feature. YAGNI principled approach here.

Lincoln-GR (approved):

PR #228, opened 2024-10-03
BDRSPS-857 Schema update for site visist data template v2

Lincoln-GR:
* fixed bug where `unique` constraints in our schema files weren't getting passed to frictionless
    * added test case to site data v2 to test this
* removed dev comment from schema
* allow multiple `siteID`s in site visit template

### DIFF
--- a/abis_mapping/templates/survey_site_visit_data_v2/schema.json
+++ b/abis_mapping/templates/survey_site_visit_data_v2/schema.json
@@ -113,7 +113,6 @@
 			}
 		},
 		{
-			"_comment": "can we do a flexible vocabulary while I do the recon to establish the concepts?",
 			"name": "protocolName",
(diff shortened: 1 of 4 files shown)


abis_mapping/templates/survey_site_visit_data_v2/schema.json
```diff
@@ -19,8 +19,7 @@
 			"type": "string",
 			"format": "default",
 			"constraints": {
-				"required": true,
-				"unique": true
+				"required": true
```
Lincoln-GR, line 22: @joecrowleygaia do you agree with removing `unique: true` from `siteID`? Otherwise you won't be able to have multiple visits per site in the template

joecrowleygaia (approved):

PR #230, opened 2024-10-03
Bdrsps 878: Added incidental occurrence v1 page to archived_templates

joecrowleygaia:


docs/pages/archived
Lincoln-GR, line 1: remove this?

Lincoln-GR (commented):
should we keep this on hold till incidental v1 is actually archived?

docs/pages/archived
joecrowleygaia: thought we said just push it in for now. It will only apply to the `dev` version of the instructions which shouldn't be getting used by any endusers

joecrowleygaia: oh remove the dir. Yeah it doesn't show up in my local repo any more. Not sure what the go is with it. Previous comment is regarding the on hold question.

joecrowleygaia: ok i could just remove it in remote using the ui.

Lincoln-GR: git doesn't track directories... this is an empty file somehow

Lincoln-GR:
> thought we said just push it in for now. It will only apply to the dev version of the instructions which shouldn't be getting used by any endusers

yes but it will go to staging and uat when we next tag

docs/pages/archived
joecrowleygaia: it was a directory

joecrowleygaia:
> > thought we said just push it in for now. It will only apply to the dev version of the instructions which shouldn't be getting used by any endusers
> 
> yes but it will go to staging and uat when we next tag

@serge-gaia are you concerned?

serge-gaia (approved):

PR #234, opened 2024-10-03
Lint unused suppressions

joecrowleygaia:
Was getting twitchy seeing all these unused linter suppressions.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -7,3 +7,4 @@
 
 # 2024-10-03 Applied ruff check --fix-only with added RUF100 check (unused suppressions)
 7b6622528ec10913c7ef169c3089a38697ca766c
+4f215c88d4e1532e74bfbce5b7c4c6a03bee7ccb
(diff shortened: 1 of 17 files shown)


.git-blame-ignore-revs
```diff
@@ -4,3 +4,7 @@
 
 # 2024-10-03 Applied ruff format to all python code
 98477463d23e270c276097edde9eb41a8f944ca2
+
+# 2024-10-03 Applied ruff check --fix-only with added RUF100 check (unused suppressions)
+7b6622528ec10913c7ef169c3089a38697ca766c
+4f215c88d4e1532e74bfbce5b7c4c6a03bee7ccb
```
Lincoln-GR, line 10: This fixup commit only touches one file. I don't think we should set the precedent of adding such commits to this file.

Either squash the fixup commit or accept it will appear in blame

Lincoln-GR (commented):
good rule to enable :+1: I have one opinion on the second ignored commit

.git-blame-ignore-revs
joecrowleygaia: I believe anyone that does a small autoformat specifically on a file they didn't interact with should absolutely add it to blame ignore. Closing PR and deleting branch to squash.

PR #241, opened 2024-10-04
BDRSPS-845 Updated survey occurrence v2 rdf mapping as required.

joecrowleygaia:
Mostly copied over from incidental template where possible in addition to template specific update.


abis_mapping/templates/survey_occurrence_data_v2/mapping.py
```diff
@@ -457,6 +461,18 @@ def apply_mapping_row(
...
 
+        # Conditionally create uri dependent on surveyID field.
+        if survey_id := row["surveyID"]:
+            survey = utils.rdf.uri(f"survey/{urllib.parse.quote(survey_id, safe='')}", base_iri)
+        else:
+            survey = None
+
+        # Conditionally create uri dependent on siteVisitID field.
+        if site_visit_id := row["siteVisitID"]:
+            site_visit = utils.rdf.uri(f"visit/{urllib.parse.quote(site_visit_id, safe='')}", base_iri)
```
Lincoln-GR, line 472: Is there a reason these two new uris (`site_visit` and `survey`) are using `quote()` on the value from the row but most of the others generated above don't? - just thinking it would be nice to have a consistent approach if possible.

Also because `utils.rdf.uri()` uses slugify internally any url-quoted chars have their `%` changed to `-`, e.g.
```python
site_visit_id = "A11/22"  # will result in:
site_visit = rdflib.term.URIRef('http://createme.org/visit/A11-2F22')
```
Note sure if this really matters?

Lincoln-GR (commented):
Looks good, one question/comment

abis_mapping/templates/survey_occurrence_data_v2/mapping.py
joecrowleygaia: Yeah you're right to flag it. It has kept me awake at night. The case for doing it this way for IDs was to capture the value in the uri since it isn't being recorded in a string literal in the graph, which is commonly done for the slugified uris. This was the compromise that was reached, not really what the uri should be used for but is a stop gap for now. 
As an aside rdf.uri() retains forward slashes currently so that result there should be
```
site_visit_id = "A11/22"  # will result in:
site_visit = rdflib.term.URIRef('http://createme.org/visit/A11/2F22')
```
Let me know if it isn't as that could be a bug.

Lincoln-GR: I'm pretty sure my result is correct since `urllib.parse.quote(...,  safe="")` removes the forward slashes before `rdf.uri()` gets it.
i.e.
```python
urllib.parse.quote("A11/22", safe='') == "A11%2F22"
utils.rdf.uri(f"visit/A11%2F22") == rdflib.term.URIRef('http://createme.org/visit/A11-2F22')
```
I think you could change it to `urllib.parse.quote(...,  safe="/")` to retain the forwards slashes.

BUT if the idea is to capture the original value in the URI then I think any forwards slashes from the value SHOULD be quoted, so they don't become a "separator" in the URI. 

<hr>

Ok it the idea is to capture the values in the UIRs that sounds ok.
The problem I think is that any url-quoted chars from `quote()` are getting mangled by slugify in `rdf.uri()` (`%` -> `-`)
That would make it difficult for anyone to un-quote and get the original value.
Can we turn off the slugifying in this case, since we've already url-quoted the input data, should be safe to do so.

Lincoln-GR: i.e. I'm thinking the ideal result would be
```python
site_visit_id = "A11/33+"  
site_visit = rdflib.term.URIRef("http://createme.org/visit/A11%2F33%2B")
#                                                          ^^^^^^^^^^^ 
#                  This part is the original ID url-quoted and without any slugifying
```

joecrowleygaia: change is in

Lincoln-GR (approved):

PR #257, opened 2024-10-10
[BDRSPS-896] Added `template_status` to template metadata schema

chungvl:


abis_mapping/types/metadata.py
```diff
@@ -23,6 +23,7 @@ class TemplateMetadata(pydantic.BaseModel):
     sampling_type: str
     template_url: str
     schema_url: str
+    template_status: str
```
serge-gaia, line 26: I'm trying to find a better name, like 'lifecycle_status` or `support_status`.
Also I think the type should be an `Enum`, with the values:
* "Beta"
* "Current"
* "Deprecated"

serge-gaia: I opt for `lifecycle_status`

chungvl: The phrase `lifecycle` is going to get confusing given we're already talking about biology?

abis_mapping/types/metadata.py
```diff
@@ -23,6 +24,7 @@ class TemplateMetadata(pydantic.BaseModel):
     sampling_type: str
     template_url: str
     schema_url: str
+    template_lifecycle_status: Literal["Beta", "Current", "Deprecated"]
```
serge-gaia, line 27: An `Enum` would have been good here, I think.

serge-gaia: But that's fine, merging.

serge-gaia (approved):

PR #258, opened 2024-10-11
BDRSPS-909 Cache the creation of pyproj objects

Lincoln-GR:
As per pyproj docs: https://pyproj4.github.io/pyproj/stable/advanced_examples.html#caching-pyproj-objects.

I'm using `functools.cache` instead of `functools.lru_cache` because there are only so many datums, so the cache can only grow so large, so we don't need the "lru" eviction logic.

### DIFF
--- a/abis_mapping/types/spatial.py
+++ b/abis_mapping/types/spatial.py
@@ -2,6 +2,7 @@
 
 # Standard
 import decimal
+import functools
 import re
 
 # Third-party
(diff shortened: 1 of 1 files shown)


serge-gaia:
Noice!

abis_mapping/types/spatial.py
```diff
@@ -19,6 +20,12 @@
 from typing import NamedTuple
 
 
+# Create cached lookup functions for pyproj.CRS and pyproj.Transformer objects,
+# to avoid the cost of repeatedly creating them when mapping.
+_CRS_cached = functools.cache(pyproj.CRS)
+_transformer_from_crs_cached = functools.cache(pyproj.Transformer.from_crs)
```
serge-gaia, line 26: :+1:

serge-gaia (approved):

PR #260, opened 2024-10-15
BDRSPS-900 Archive incidental template v1

Lincoln-GR:
Archiving a template makes it no longer available. 
The incidental v1 docs are already in the archived section.
The tests I deleted were specific only to v1, don't need to be adjusted to v2.


joecrowleygaia (commented):
There should be a scenario (or scenarios) in the `tests/templates/conftest.py` to be removed.

joecrowleygaia (approved):

PR #269, opened 2024-10-21
BDRSPS-916 Temporal flexibility occurrence side implementation ssd. 

joecrowleygaia:
- Site visit id cross validation now added to occurrence template
- Survey occurrence mapping now takes site visit id to default temporal entity rdf map.
- An attempt at implementing a way to prevent name collisions of rdflib blank nodes when performing set theoretic union operations is included, but not used.

### DIFF
--- a/abis_mapping/templates/survey_occurrence_data_v2/mapping.py
+++ b/abis_mapping/templates/survey_occurrence_data_v2/mapping.py
@@ -280,7 +280,7 @@ class SurveyOccurrenceMapper(base.mapper.ABISMapper):
                 ingest before yielding a graph. `None` will ingest all rows.
             site_id_geometry_map (dict[str, str]): Default values of geometry wkt
                 to use for a given site id.
-            site_viset_id_temporal_map (dict[str, str]): Default values of
+            site_visit_id_temporal_map (dict[str, str]): Default values of
                 temporal entity rdf, as turtle, to use for a given site visit id.
 
         Yields:
--- a/abis_mapping/templates/survey_occurrence_data_v2/mapping.py
+++ b/abis_mapping/templates/survey_occurrence_data_v2/mapping.py
@@ -6,6 +6,7 @@ import urllib.parse
 # Third-Party
 import frictionless
 import rdflib
+import rdflib.term
 
 # Local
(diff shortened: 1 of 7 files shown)


tests/utils/test_terms.py
```diff
@@ -0,0 +1,89 @@
...
+@pytest.fixture(scope="module")
+def patch_rdflib_bnode() -> Iterable[None]:
+    """Expected overriding/patching of rdflib.BNode will be done at the start of a module"""
+    rdflib.BNode = utils.terms.BNodeAlwaysUUID4  # type: ignore[misc]
+    rdflib.term.BNode = utils.terms.BNodeAlwaysUUID4  # type: ignore[misc]
+
+    yield None
+
+    rdflib.BNode = rdflib.BNode  # type: ignore[misc]
+    rdflib.term.BNode = rdflib.term.BNode  # type: ignore[misc]
```
Lincoln-GR, line 30: I think this may not be un-patching as intended... `rdflib.BNode` is now the custom class, so this is a no-op

Just use the regular patcher e.g. `mocker.patch.object(rdflib, "BNode", new=utils.terms.BNodeAlwaysUUID4`

abis_mapping/templates/survey_occurrence_data_v2/mapping.py
```diff
@@ -262,6 +280,8 @@ def apply_mapping(
                 ingest before yielding a graph. `None` will ingest all rows.
             site_id_geometry_map (dict[str, str]): Default values of geometry wkt
                 to use for a given site id.
+            site_viset_id_temporal_map (dict[str, str]): Default values of
```
Lincoln-GR, line 283: `viset` typo

tests/utils/test_terms.py
joecrowleygaia: fixed.

abis_mapping/templates/survey_occurrence_data_v2/mapping.py
joecrowleygaia: fixed.

Lincoln-GR (approved):

PR #273, opened 2024-10-22
BDRSPS-872 Rectified issue of incorrect xy placement of coordinates when generating rdf wkt literals with crs defined.

joecrowleygaia:
The coordinates are stored internally within the `Geometry` object as long-lat up until rdf wkt literal output is to occur. The same is true for reading from a geosparql string in reverse.

### DIFF
--- a/abis_mapping/types/spatial.py
+++ b/abis_mapping/types/spatial.py
@@ -8,6 +8,7 @@ import re
 # Third-party
 import shapely
 import shapely.ops
+import numpy as np
 import pyproj
 import rdflib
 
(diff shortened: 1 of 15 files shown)


abis_mapping/types/spatial.py
```diff
@@ -8,6 +8,7 @@
 # Third-party
 import shapely
 import shapely.ops
+import numpy as np
```
Lincoln-GR, line 11: Add numpy to `[tool.poetry.dependencies]` in pyproject.toml since we are adding a direct import of it - 
`$ poetry add "numpy^1.0.0"`

Lincoln-GR (changes requested):
Looks good, one thing to address

abis_mapping/types/spatial.py
joecrowleygaia: yeah i wasn't sure on this one since it is closely tied to shapely. Have added it.

Lincoln-GR (approved):

PR #279, opened 2024-10-24
BDRSPS-925 Add Collections for certain attributes in the survey occurrence v2 mapping

Lincoln-GR:
BDRSPS-925 Add Collections for certain attributes in the survey occurrence v2 mapping
    
The Attribute and Attribute Value IRIs now have the value in them, instead of the row number.
This greatly reduces the number we create.
A new Collection node joins the Attribute node and the rows that had that attribute.
    
The following attributes are changed:
* basisOfRecord
* habitat
* preparations
* dataGeneralizations
    
Same as #272, but for the survey occurrence template

### DIFF
--- a/abis_mapping/templates/survey_occurrence_data_v2/examples/margaret_river_flora/margaret_river_flora.ttl
+++ b/abis_mapping/templates/survey_occurrence_data_v2/examples/margaret_river_flora/margaret_river_flora.ttl
@@ -23,6 +23,62 @@
     tern:hasSimpleValue "Category 1 - Department of Biodiversity and Conservation" ;
     tern:hasValue <http://createme.org/value/sensitivityCategory/16> .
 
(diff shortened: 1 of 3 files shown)


joecrowleygaia (commented):
Was omission of the types in the docstrings on purpose? happy to approve if so (it's not a requirement anywhere, just what was done before by @HaydenGaia in the past i think)

Lincoln-GR:
> Was omission of the types in the docstrings on purpose? happy to approve if so (it's not a requirement anywhere, just what was done before by @HaydenGaia in the past i think)

Yes the [google style guide](https://google.github.io/styleguide/pyguide.html#doc-function-args) says you can leave out the types from the docstring when the code has type annotations.

joecrowleygaia (approved):

PR #286, opened 2024-10-29
BDRSPS-953 Added chained inclusion check to siteVisitID and siteID for the survey occurrence v2 template.

joecrowleygaia:
Adds a new custom check `ChainedInclusion`


abis_mapping/plugins/chained_inclusion.py
```diff
@@ -0,0 +1,61 @@
...
+
+        # Check chained inclusion
+        # missing must be empty
+        if not missing:
+            # Valid
+            return
+
+        note = (
+            f"the columns {self.field_names} are chained inclusive and values"
+            " must be provided together if its preceding field contains a value"
```
Lincoln-GR, line 53: Suggest `f"The column(s) {self.field_names[1:]} must be provided when the column {self.field_names[0]} is provided."`

Lincoln-GR (commented):
Looks good, one suggestion

abis_mapping/plugins/chained_inclusion.py
joecrowleygaia: "me fail english? That's umpossible."
Changed. Much better.

Lincoln-GR (approved):

PR #288, opened 2024-10-30
BDRSPS-954 Created siteVisitID -> siteID map extraction method for the survey site visit template.

joecrowleygaia:


abis_mapping/templates/survey_site_visit_data_v2/mapping.py
```diff
@@ -112,6 +112,37 @@ def apply_validation(
         # Return validation report
         return report
 
+    def extract_site_id_map(
```
Lincoln-GR, line 115: "site_id_map" is already a thing from one of the other extraction methods.
Can this be `extract_site_visit_id_to_site_id_map` to avoid confusion

Lincoln-GR (commented):
Nice, one comment

abis_mapping/templates/survey_site_visit_data_v2/mapping.py
joecrowleygaia: done.

Lincoln-GR (approved):

PR #290, opened 2024-10-30
[BDRSPS-918] First stab adding `coordinateUncertaintyInMeters`

chungvl:


abis_mapping/templates/survey_metadata/mapping.py
```diff
@@ -520,7 +520,8 @@ def add_spatial_coverage(
         graph.add((geometry_node, utils.namespaces.GEO.asWKT, geometry.to_transformed_crs_rdf_literal()))
 
         if spatial_accuracy is not None:
-            graph.add((geometry_node, utils.namespaces.GEO.hasMetricSpatialAccuracy, spatial_accuracy))
+            accuracy = rdflib.Literal(spatial_accuracy, datatype=rdflib.XSD.double)
+            graph.add((geometry_node, utils.namespaces.GEO.hasMetricSpatialAccuracy, accuracy))
```
serge-gaia, line 524: looks good

abis_mapping/templates/survey_metadata/mapping.py
```diff
@@ -502,6 +502,7 @@ def add_spatial_coverage(
         # Extract relevant values
         datum = row["geodeticDatum"]
         sc_geometry = row["spatialCoverageWKT"]
+        spatial_accuracy = row["coordinateUncertaintyInMeters"] if "coordinateUncertaintyInMeters" in row else None
```
joecrowleygaia, line 505: could use `spatial_accuracy = row.get("coordinateUncertaintyInMeters")`

abis_mapping/templates/survey_metadata/mapping.py
joecrowleygaia: looks like the way to go when adding the accuracy to the graph.

joecrowleygaia (commented):
I'm not sure that it is required for metadata template since the field doesn't exist.

abis_mapping/templates/survey_metadata/mapping.py
```diff
@@ -518,6 +519,10 @@ def add_spatial_coverage(
         graph.add((geometry_node, a, utils.namespaces.GEO.Geometry))
         graph.add((geometry_node, utils.namespaces.GEO.asWKT, geometry.to_transformed_crs_rdf_literal()))
 
+        if spatial_accuracy is not None:
+            accuracy = rdflib.Literal(spatial_accuracy, datatype=rdflib.XSD.double)
+            graph.add((geometry_node, utils.namespaces.GEO.hasMetricSpatialAccuracy, accuracy))
+
         self.add_geometry_supplied_as(
             subj=uri,
             pred=utils.namespaces.GEO.hasGeometry,
```
joecrowleygaia, line 528: will need to add arg `spatial_accuracy=accuracy` to this call i think and default `accuracy` to `None` i.e. `accuracy: rdflib.Literal | None = None`

chungvl: If accuracy isn't available, wouldn't you just not create the node?

joecrowleygaia: can't have it both ways :smile: ! If you're going to add the code for spatial accuracy since it doesn't exist as a field yet in the template, then you'll need to make sure the rest flows through, including a "geometry supplied as" node using the `add_geometry_supplied_as` method with the spatial accuracy provided to it.

abis_mapping/templates/survey_occurrence_data_v2/mapping.py
joecrowleygaia, line 1: same comments as incidental occurrence mapping.py changes applies here.

joecrowleygaia (changes requested):

abis_mapping/templates/incidental_occurrence_data_v3/mapping.py
```diff
@@ -1441,6 +1441,11 @@ def add_sampling_field(
         graph.add((geometry_node, a, utils.namespaces.GEO.Geometry))
         graph.add((geometry_node, utils.namespaces.GEO.asWKT, geometry.to_transformed_crs_rdf_literal()))
 
+        spatial_accuracy = row.get("coordinateUncertaintyInMeters")
```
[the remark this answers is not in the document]
serge-gaia: you're sure of that @joecrowleygaia ? I prefer to deal with `None` that with an exception.  
Open discussion.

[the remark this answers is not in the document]
serge-gaia: but we've done with direct key access so far. So let be consistent.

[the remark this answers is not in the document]
serge-gaia: After reflection, you're right @joecrowleygaia. Better to use direct key access to early detect schema changes.

Lincoln-GR:
@chungvl Can you try running the `poe generate-example-ttl-files` script and committing the result to this PR.
This should greatly reduce the overall diff the `.ttl` files to make them easier to review, and surface the actual changes you are making.

chungvl:
```Traceback (most recent call last):
  File "/home/voonli.chung/PycharmProjects/abis-mapping/./scripts/generate_example_ttl_files.py", line 120, in <module>
    main()
  File "/home/voonli.chung/PycharmProjects/abis-mapping/./scripts/generate_example_ttl_files.py", line 56, in main
    raise RuntimeError("BNode.__new__ had unexpected signature, the following code might need to be updated")
RuntimeError: BNode.__new__ had unexpected signature, the following code might need to be updated
```

abis_mapping/templates/incidental_occurrence_data_v3/mapping.py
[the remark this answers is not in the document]
chungvl: I'll move it back to the way it was before it was suggested I use `get` in a previous comment.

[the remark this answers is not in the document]
joecrowleygaia: If that is relating to the one i suggested changing to `get`, it was slightly different since it was accessing a field that didn't exist on the template. Sorry for the confusion.

poetry.lock
joecrowleygaia, line 1: I think this means there was an upgrade applied to the poetry project, or the lockfile was deleted and recreated. Does it need to be left as-was @serge-gaia or @Lincoln-GR ?

Lincoln-GR: Looks like the sub-dependencies were bumped. This will only effect development and CI on this project, not part-a.
Should be fine to leave it. Or on the other hand no reason it needs to be in this PR :shrug:

chungvl: Prior to doing an update the `generate-example-ttl` script wasn't working for me.

joecrowleygaia (approved):

poetry.lock
Lincoln-GR: For next time, I would imagine just doing `poetry install` to install the current dependencies, should be sufficient to get the script working without changing the lock file.
(Assuming there isn't an actual incompatibility between our code and a dependency)

PR #295, opened 2024-10-31
BDRSPS-951 Updated the mapping for attributes for the plan node of the metadata template.

joecrowleygaia:
This will remove the creation of CREATEME iri from within the template's mapping methods. 
The attributes being assigned previously were incorrect and now use those IRIs created for `tern:Attribute` nodes.

### DIFF
--- a/abis_mapping/templates/survey_metadata_v2/mapping.py
+++ b/abis_mapping/templates/survey_metadata_v2/mapping.py
@@ -692,7 +692,7 @@ class SurveyMetadataMapper(base.mapper.ABISMapper):
 
         Args:
             uri: Plan reference.
-            survey_type_attribute: SurveyType attribute for the node 
+            survey_type_attribute: SurveyType attribute for the node
             target_habitat_scope_attribute: targetHabitatScope attribute for the node.
             target_taxa_attribute: target taxa attribute for the node.
             row: Raw data row.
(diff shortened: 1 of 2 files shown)


abis_mapping/templates/survey_metadata_v2/mapping.py
```diff
@@ -681,44 +682,34 @@ def add_agent(
...
-                )
-            )
+        if survey_type_attribute:
+            graph.add((uri, utils.namespaces.TERN.hasAttribute, survey_type_attribute))
+
+        for hbt_attr in target_habitat_scope_attributes:
+            graph.add((uri, utils.namespaces.TERN.hasAttribute, hbt_attr))
+
+        for tx_attr in target_taxa_attributes:
+            graph.add((uri, utils.namespaces.TERN.hasAttribute, tx_attr))
```
Lincoln-GR, line 712: I agree with removing the old code, not sure we need this new stuff.
These 3 attributes have collections I added in #287.
The attribute nodes are linked to the collections. Then the collections are linked to the `tern:Survey` node.
Not sure if we also need a link from the `prov:Plan` to the attributes themselves? Or should the `prov:Plan` link to the collections?

joecrowleygaia: no idea what any of that means. The scope for me here was to clean up a section of mapping code that was creating CREATEME IRIs incorrectly. I think the validity of the node's shape can be handled in subsequent PRs.

joecrowleygaia: oh sorry i dont see it as new. It's replacing.

Lincoln-GR (approved):

PR #297, opened 2024-11-01
Add script for re-generating example ttl files

Lincoln-GR:


scripts/generate_example_ttl_files.py
```diff
@@ -0,0 +1,120 @@
...
+    for frame in inspect.stack():
+        # NOTE "abis_mapping" with underscore is the module name.
+        # NOT "abis-mapping" with a dash which is the repository name.
+        if "abis_mapping" in frame.filename:
+            code_path_hash.update(frame.function.encode("utf-8"))
+            # we don't include the frame.filename in the hash because this wil be
+            # different depending on where abis-mapping is checked out.
+    digest = code_path_hash.digest()
+    number = next(COUNTERS[digest]).to_bytes(4, "big")
+    return (digest + number).hex()  # concatenate bytes and return hex representation
```
joecrowleygaia, line 116: not sure this level of name creation is needed. I think all that is required is that the ID starts with a letter (in rdflib they use "N") but after that can use alphanumerics and a few select symbols. Why not just make the id a sequential number generator with a prefix of "L". A key thing to note is that [blank nodes don't resolve outside of a file](https://www.w3.org/TR/rdf12-concepts/#section-blank-nodes).

scripts/generate_example_ttl_files.py
```diff
@@ -0,0 +1,120 @@
...
+        raise RuntimeError("BNode.__new__ had unexpected signature, the following code might need to be updated")
+    # Override default _sn_gen function with our custom one.
+    rdflib.BNode.__new__.__defaults__ = (defaults[0], _custom_sn_gen, defaults[2])
+
+    # 2. second override:
+    # Mock "Date.today()" function to always give the same result
+    unittest.mock.patch.object(
+        abis_mapping.types.temporal.Date,
+        "today",
+        new=lambda: abis_mapping.types.temporal.Date(2020, 1, 1),
```
joecrowleygaia, line 65: please set to your birthday or another date of significance to you :stuck_out_tongue_winking_eye:

joecrowleygaia (commented):
Simplify the blank node identifier number generation please and set prefix to "L"

scripts/generate_example_ttl_files.py
Lincoln-GR: "L" for Lincoln? :laughing: 

I didn't use a simple sequential number generator because then a mapping change which changes the number of blank nodes created, will change the IDs of all following B nodes.
The point of this (I admit complex) generation, is to try to have the B Node IDs change as little as possible, even when mapping is changed, so the git diff is reduced

serge-gaia (approved):

scripts/generate_example_ttl_files.py
```diff
@@ -0,0 +1,120 @@
...
+    # i.e. when the same hash occurs, they will be numbered 0,1,2,etc. at the end of the ID.
+    # This ensures each ID should be globally unique.
+    code_path_hash = hashlib.blake2b(digest_size=12, person=b"gen_b_node_id")
+    # hash ttl file path.
+    code_path_hash.update(TTL_FILE_PATH.encode("utf-8"))
+    # Hash functions in the abis_mapping module called to make this BNode()
+    for frame in inspect.stack():
+        # NOTE "abis_mapping" with underscore is the module name.
+        # NOT "abis-mapping" with a dash which is the repository name.
+        if "abis_mapping" in frame.filename:
```
joecrowleygaia, line 110: Looking at the output turtle, i don't think this is retrieving the correct frame. I think you would rather inspect the frame.function and look out for `apply_mapping` (or both?)? Again though i don't think you need this at all. A single counter that gets reset each mapping file should be enough, also making this function part of a closure would be a better practice rather than relying on globals for it i think.

abis_mapping/templates/survey_site_data_v2/examples/minimal.ttl
```diff
@@ -85,47 +85,47 @@
...
     rdf:subject <http://createme.org/dataset/Example-Systematic-Survey-Site-Dataset/Site/P1> ;
     rdfs:comment "supplied as" .
 
-_:N1c3ebed8bb624c9ba69063ad0c274e88 a geo:Geometry ;
+_:N2d5e0873e8e72e68e4a9793300000000 a geo:Geometry ;
     geo:asWKT "<http://www.opengis.net/def/crs/EPSG/0/7844> POINT (-33.85 114.99)"^^geo:wktLiteral ;
     geo:hasMetricSpatialAccuracy 5e+01 .
 
-_:N2952b6066f2b4cebb9b4a50d41ebb8ea a geo:Geometry ;
+_:N53dd57c04e35e4aa3188c0b800000000 a geo:Geometry ;
```
joecrowleygaia, line 124: see my comment surrounding the searching for the correct frame.  I don't believe these IDs should be different following the N prefix up until their last 1 or 2 digits.

Lincoln-GR: The first ID is for a Bnode created in `add_point_geometry`,
the second is for one created in `add_footprint_geometry`,
those functions names are included in the hash, so I would expect these IDs to be different in the hash part of the ID

scripts/generate_example_ttl_files.py
Lincoln-GR: Multiple frames' `function` are being included in each hash -
from my debugging the effective list of `frame.function`s included in each hash are things like
```python
['original_datum_uri', 'to_rdf_literal', 'add_geometry_supplied_as', 'add_sampling_field', 'apply_mapping_row', 'apply_mapping']
['add_observation_scientific_name', 'apply_mapping_row', 'apply_mapping']
['add_observation_verbatim_id', 'apply_mapping_row', 'apply_mapping']
```
So BNodes created from different parts of the mapping code have different hashes

Lincoln-GR: This `if` check is really just to exclude the inner part of the stack - which is rdflib internals, 
and to exclude the outer part of the stack (everything before `apply_mapping`),
which could be different depending on if you're running the script directly, or from a debugger or something

scripts/generate_example_ttl_files.py
joecrowleygaia: Pretty certain the "N" is for Nick. Solid reasoning for going this route.

joecrowleygaia (approved):

PR #305, opened 2024-11-06
BDRSPS-940 Adding dwc:Occurrence and abis:BiodiversityRecord nodes to occurrence graphs

joecrowleygaia:
- Removes the sample field and sampling field nodes and combines this within dwc:Occurrence node. 
- Removes the terminal foi as part of this. 
- Changed the constraints on both the occurrence `providerRecordID` fields to be unique.
- Attempted realignment of site visit, site and survey IRI construction between the 4 ssd templates using the spec provided as part of this ticket. This is likely going to be protected against somehow in the future.


abis_mapping/templates/survey_occurrence_data_v2/mapping.py
```diff
@@ -1560,6 +1554,9 @@ def add_observation_verbatim_id(
         graph.add((uri, rdflib.SOSA.hasResult, verbatim_id))
         graph.add((uri, rdflib.SOSA.hasSimpleResult, rdflib.Literal(row["verbatimIdentification"])))
         graph.add((uri, rdflib.SOSA.observedProperty, CONCEPT_TAXON))
+
+        # Declare temporal entity allowing for correct type assignments
+        temporal_entity: rdflib.term.Node | None = None
```
Lincoln-GR, line 1559: when you are declaring a variable mainly to give it a certain type, 
and the variable is given a value in every following code path,
you can leave out the initial value, e.g.
```python
temporal_entity: rdflib.term.Node | None  # leave out = None
```
no need to change this, just a pro tip

abis_mapping/templates/survey_occurrence_data_v2/mapping.py
```diff
@@ -4498,8 +4305,178 @@ def add_sensitivity_category_collection(
         # Add link to attribute
         if sensitivity_category_attribute:
             graph.add((uri, utils.namespaces.TERN.hasAttribute, sensitivity_category_attribute))
-        # TODO add link to the abis:BiodiversityRecord node once that is implemented
-        # graph.add((uri, rdflib.SDO.member, ...))
+        # Add link to the biodiversity record
+        graph.add((uri, rdflib.SDO.member, provider_record_id_biodiversity_record))
```
Lincoln-GR, line 4309: I love it when a plan comes together

Lincoln-GR (approved):
great work!

abis_mapping/templates/survey_occurrence_data_v2/mapping.py
joecrowleygaia: yeah i know, for whatever reason I decided to go this way.

PR #307, opened 2024-11-07
[BDRSPS-986] Updated template statuses

chungvl:


serge-gaia:
I'm keeping this PR until we are close to production release.

serge-gaia:
@Lincoln-GR , when we merge that do we have to do some manual documentation update to reflect the status changes or everything is automatic?

Lincoln-GR:
Everything should be automatic

PR #310, opened 2024-11-11
BDRSPS-819 Made necessary tests or changes to types code after investigating low coverage on tests.

joecrowleygaia:
Probably the biggest change is no longer returning None when retrieving datum uris. The thought here is that previously when these were directly retrieved from within the mapping methods, that there were no protections against the vocab error that would be thrown as well as the fact that the data that is expected to hit this method is already expected to have been validated through the enum check of the `geodeticDatum` fields. A chained exception is raised in its place.

Also renamed non-uppercase comment "note" with "NOTE" to help IDEs identify better.

### DIFF
--- a/abis_mapping/types/spatial.py
+++ b/abis_mapping/types/spatial.py
@@ -112,8 +112,7 @@ class Geometry:
             return vocab(graph=rdflib.Graph()).get(self.original_datum_name)
         except utils.vocabs.VocabularyError as exc:
             raise GeometryError(
-                f"CRS {self.original_datum_name} is "
-                "not defined for the GEODETIC_DATUM fixed vocabulary."
+                f"CRS {self.original_datum_name} is " "not defined for the GEODETIC_DATUM fixed vocabulary."
             ) from exc
 
     @property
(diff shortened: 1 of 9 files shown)


abis_mapping/types/spatial.py
```diff
@@ -100,14 +99,21 @@ def original_datum_uri(self) -> rdflib.URIRef | None:
         """Getter for the original datum URI.
 
         Returns:
-            rdflib.URIRef: Uri corresponding to original datum if known, else None.
+            rdflib.URIRef: Uri corresponding to original datum.
```
Lincoln-GR, line 102: update function return type to just `rdflib.URIRef`

abis_mapping/types/spatial.py
```diff
@@ -126,15 +132,23 @@ def transformer_datum_uri(self) -> rdflib.URIRef | None:
...
+        vocab = utils.vocabs.get_vocab("GEODETIC_DATUM")
+        default_crs = settings.Settings().DEFAULT_TARGET_CRS
 
-        # If vocab doesn't exist
-        return None
+        try:
+            # Init with dummy graph and return corresponding uri
+            return vocab(graph=rdflib.Graph()).get(default_crs)
+        except utils.vocabs.VocabularyError as exc:
+            raise GeometryError(
```
Lincoln-GR, line 149: while you're here can you change `GeometryError` class to inherit from `Exception` not `BaseException`

tests/types/test_spatial.py
```diff
@@ -132,6 +133,85 @@ def test_geometry_transformer_datum_uri() -> None:
...
+        # Create a stubbed settings model
+        class TempSettings(settings.Settings):
+            # Modified fields below
+            DEFAULT_TARGET_CRS: str = value
+
+        # Patch Settings
+        mocker.patch(
+            "abis_mapping.settings.Settings",
+            new=TempSettings,
+        )
```
Lincoln-GR, line 166: If you change spatial.py to always look up `settings.SETTINGS.DEFAULT_TARGET_CRS` (not `settings.Settings()`),
this part of the mocking could be removed.
I'll leave this up to you.

joecrowleygaia: leave

abis_mapping/types/spatial.py
joecrowleygaia: thinking you meant change the function return *annotation*? changed.

PR #312, opened 2024-11-11
BDRSPS-966 SSD Site and SiteVisit IRI construction

Lincoln-GR:
Add pattern functions for constructing Site and Site Visit IRIs.
This should be the last changes required in the survey occurrence template

### DIFF
--- a/abis_mapping/templates/survey_occurrence_data_v2/mapping.py
+++ b/abis_mapping/templates/survey_occurrence_data_v2/mapping.py
@@ -1,8 +1,5 @@
 """Provides ABIS Mapper for `survey_occurrence_data.csv` Template v2"""
 
-# Standard
-import urllib.parse
-
 # Third-Party
 import frictionless
 import rdflib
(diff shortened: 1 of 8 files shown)


abis_mapping/templates/survey_occurrence_data_v2/mapping.py
```diff
@@ -647,15 +638,25 @@ def apply_mapping_row(
...
 
+        # Create URIs for Survey-related fields (i.e. fields not on the incidental template)
+
         # Create TERN survey IRI from surveyID field
         survey_id: str | None = row["surveyID"]
         survey = utils.iri_patterns.survey_iri(base_iri, survey_id)
 
+        # Create Tern Site IRI, depending on the siteID field
+        site_id: str | None = row["siteID"]
+        if site_id:
```
joecrowleygaia, line 649: not a walrus enjoyer?

joecrowleygaia (approved):

abis_mapping/templates/survey_occurrence_data_v2/mapping.py
Lincoln-GR: sometimes yes, but here the walrus was preventing me from adding a proper type annotation to the variables like `site_id`

joecrowleygaia: i probs didn't notice its use somewhere mypy complained.

Lincoln-GR: Well without the type annotation `row["field"]` is `Any`, and `Any` means "don't type check this", so mypy generally won't complain about that

PR #316, opened 2024-11-12
BDRSPS-980 Added model documentation check to ci

joecrowleygaia:
Made all jobs depend on `poetry-lock` due to a test that revealed all would fail on a common issue of lockfile out of sync with pyproject.toml. Added env variables for poetry and python versions.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -31,7 +31,7 @@ jobs:
         run: |
           echo "Checking if locking dependencies has created any diff in the lock file..."
           if 
-            git diff --exit-code -- poetry.lock 
+            git diff --exit-code -- poetry.lock
           then
             echo "No changes to poetry.lock detected"
           else
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -3,9 +3,13 @@ on:
   push:
     branches:
       - main
+      - BDRSPS-980
   pull_request:
     branches:
       - main
+env:
+  PYTHON_VERSION: "3.11"
+  POETRY_VERSION: "1.8.4"
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/testing.yml
```diff
@@ -26,42 +31,70 @@ jobs:
       - name: Check resolving dependencies did not change the lock file
         run: |
           echo "Checking if locking dependencies has created any diff in the lock file..."
-          if 
-            git diff --exit-code -- poetry.lock
-          then
-            echo "No changes to poetry.lock detected"
-          else
+          set +e
+          git diff --exit-code -- poetry.lock > /dev/null
```
Lincoln-GR, line 35: I intentionally had the git diff going to stdout so that someone reviewing a failing check can see the diff, can you restore the old code here.

.github/workflows/testing.yml
```diff
@@ -26,42 +31,70 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: ./.github/actions/setup-python-and-dependencies
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          poetry-version: ${{ env.POETRY_VERSION }}
+      - run: poetry run poe generate-model-docs
+      - name: Check model changes left undocumented
+        run: |
+          echo "Checking if model changes left undocumented"
+          set +e
```
Lincoln-GR, line 56: unsetting `e` and checking `$?` feels more complex than what we had before, why not simply do
```bash
if
    ! git diff --exit-code -- docs/* > /dev/null
else
    echo "changes..."
    exit 1
fi
```

Lincoln-GR (approved):

PR #322, opened 2024-11-13
BDRSPS-940 Fixup: Remove the isSampleOf predicate for sites

Lincoln-GR:
Previously `<http://createme.org/location/Australia>` was the terminal featureOfInterest. This was removed when we moved to abis:BiodiversityRecord and dwc:Occurrence in #305. This link should have been removed at that time.


joecrowleygaia (changes requested):
It isn't asked for within the scope of BDRSPS-940. This is a different property path on a different template. Raise a new ticket to get confirmation for change.

PR #324, opened 2024-11-14
BDRSPS-1016 Make siteVisitStart field mandatory for site visit template

Lincoln-GR:
This has flow-on effects;
* Remove OR check for ["siteVisitStart", "siteVisitEnd"]. This is redundant since siteVisitStart is always present.
* Remove cross-validation check for "siteVisitStart OR siteVisitID in occurrence template". This is redundant since siteVisitStart is always present.


joecrowleygaia (approved):
Nice

PR #327, opened 2024-11-15
BDRSPS-1024 Enable mypy disallow_any_generics

Lincoln-GR:
Enable the `disallow_any_generics` mypy option. 
This forces generic classes like `list` to be filled out, e.g. `list[str]`.

Note this PR only changes type hints, not the runtime code.

Without this setting, mypy treats an unfilled generic like `list` as `list[Any]`, where `Any` is the special type meaning "don't type check".
So enabling this setting and filling out all the generics, increases the amount of code that is actually type checked by mypy.


joecrowleygaia (commented):
for me using `object` is so close to using `Any` anyways. May as well just use `Any` in those circumstances, a little bit more readable for me and maybe a little bit more performant for the typechecker (as opposed to confirming an object has been supplied, pretty much anything). I might be wrong about how the typechecker handles `object` tho.

Lincoln-GR:
mypy [recommends](https://mypy.readthedocs.io/en/stable/dynamic_typing.html#any-vs-object) using `object` over `Any`

> If you’re not sure whether you need to use [object](https://docs.python.org/3/library/functions.html#object) or Any, use [object](https://docs.python.org/3/library/functions.html#object) – only switch to using Any if you get a type checker complaint.

joecrowleygaia:
Yeah that's what i thought, that it would typecheck `object`, making it less performant, having to traverse ast's as opposed to not typechecking. I invoke mypy as part of my lsp configuration and it is the thing that drags my IDE down the most, so i am a little reluctant to approve. Here I am saying i am sure i want `Any` over `object`, which is inline with their recommendation.

serge-gaia:
> Yeah that's what i thought, that it would typecheck `object`, making it less performant, having to traverse ast's as opposed to not typechecking. I invoke mypy as part of my lsp configuration and it is the thing that drags my IDE down the most, so i am a little reluctant to approve. Here I am saying i am sure i want `Any` over `object`, which is inline with their recommendation.

I've tried the typecheck (with no cache `--no-incremental`) on both main and this branch and there's very little difference in execution time
* main: 12s08
* this branch 12s26
mypy is just slow. I wish ruff had a typechecker :-)

joecrowleygaia:
i know right! The case of attempting to reference a method on an `object` as opposed to an `Any` sealed it for me. Approved.

pyproject.toml
```diff
@@ -15,6 +15,7 @@ pyproj = "^3.7.0"
 pydantic = "^2.9.2"
 pydantic-settings = "^2.6.0"
 numpy = "^2.1.2"
+attrs = "^24.2.0"
```
joecrowleygaia, line 18: Oops!

joecrowleygaia (approved):

pyproject.toml
serge-gaia: Good spot @Lincoln-GR !

PR #328, opened 2024-11-18
BDRSPS-1025 Changed related site uri creation logic

joecrowleygaia:
It is now dependent on the `relationshipToRelatedSite` field and if `schema:sameAs` then the related site ID is assumed to be a valid iri.

### DIFF
--- a/abis_mapping/templates/survey_site_data_v2/examples/minimal.ttl
+++ b/abis_mapping/templates/survey_site_data_v2/examples/minimal.ttl
@@ -11,14 +11,16 @@
 <http://createme.org/SiteCollection/dataGeneralizations/Coordinates-rounded-to-the-nearest-10-km-for-conservation-concern> a schema:Collection ;
     void:inDataset <http://createme.org/dataset/Example-Systematic-Survey-Site-Dataset> ;
     schema:identifier "Site Collection - Data Generalizations - Coordinates rounded to the nearest 10 km for conservation concern" ;
-    schema:member <http://createme.org/Site/P1> ;
+    schema:member <http://createme.org/Site/P1>,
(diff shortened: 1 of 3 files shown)


abis_mapping/templates/survey_site_data_v2/examples/minimal.ttl
```diff
@@ -81,6 +83,18 @@
...
 
+<http://createme.org/Site/P2> a tern:Site ;
+    void:inDataset <http://createme.org/dataset/Example-Systematic-Survey-Site-Dataset> ;
+    geo:hasGeometry _:N2d5e0873e8e72e68e4a9793300000001,
+        _:N53dd57c04e35e4aa3188c0b800000002 ;
+    schema:additionalType <http://linked.data.gov.au/def/tern-cv/8cadf069-01d7-4420-b454-cae37740c2a2> ;
+    schema:description "Fine woody debris." ;
+    schema:identifier "P2"^^<https://linked.data.gov.au/dataset/bdr/datatypes/siteID/WAM> ;
+    schema:name "Plot 2" ;
+    schema:sameAs <http://example.com/sites/S0> ;
```
Lincoln-GR, line 94: I think the ticket was asking for this to be 
```
schema:sameAs "http://example.com/sites/S0"^^xsd:anyURI
```
Or when the related site value is not a url, just a string literal
```
schema:sameAs "APASH02"
```

Lincoln-GR (changes requested):

abis_mapping/templates/survey_site_data_v2/examples/minimal.ttl
joecrowleygaia: scrapping this whole thing and starting again.

PR #330, opened 2024-11-18
BDRSPS-1029 Refactor so that extra_fields_schema is only called once per template mapping

Lincoln-GR:
Call `extra_fields_schema(..., full_schema=False)` once per template, rather than once per row.

This is an expensive function, so this should give a noticeable performance improvement.
My testing locally shows a 13% speedup when mapping a incidental v3 template with 1000 rows.

Since all the rows of a template have the same extra fields, this is a safe refactor.


joecrowleygaia:
Makes sense. This was a leftover from when we originally just suppressed the extra-label errors per line.

joecrowleygaia (approved):

PR #331, opened 2024-11-18
BDRSPS-1025 Made related site node type and value be dependent on the value of the relationshipToRelatedSite field of the site template.

joecrowleygaia:
Applied only to the site data v2 template.


Lincoln-GR (approved):
Nice work :100:

PR #345, opened 2024-11-25
BDRSPS-1028 Add internal siteID validation for relatedSiteID when relationship is `partOf`

joecrowleygaia:
created customised row constraint check to accept side inputs and app…lied that to the validation of relatedSiteID on the site template.

Please feel free to disapprove of the approach, since i see it as a step towards going away from how we have been doing things, making a separate custom checks everytime a simple-ish row constraint validation needs to be performed. Following on from this i think some of the other checks could be decommissioned e.g. LogicalOr, DefaultLookup, Vlookup.

### DIFF
--- a/abis_mapping/templates/survey_site_data_v2/mapping.py
+++ b/abis_mapping/templates/survey_site_data_v2/mapping.py
@@ -7,6 +7,7 @@ import decimal
 # Third-party
 import rdflib
 import frictionless
+import frictionless.checks
 import shapely
 import shapely.geometry
 
@@ -18,7 +19,7 @@ from abis_mapping import utils
 from abis_mapping import vocabs
 
 # Typing
-from typing import Any, Optional, Iterator
+from typing import Any, Optional, Iterator, Literal
 
 
 # Constants and shortcuts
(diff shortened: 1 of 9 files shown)


abis_mapping/templates/survey_site_data_v2/mapping.py
```diff
@@ -101,13 +102,51 @@ def apply_validation(
...
             )
         )
 
         # Return validation report
         return report
 
+    def extract_site_ids(
+        self,
+        data: base.types.ReadableType,
+    ) -> dict[str, Literal[True]]:
```
Lincoln-GR, line 126: no need to change this I'm just curious, is there a reason we don't return `set[str]` here? is it to be consistent with the other extract methods? then my my question is why don't those return `set[str]` instead of `dict[str, True]`?

Lincoln-GR (changes requested):
Sorry I do disagree with this approach. If anything I think we should be moving in the opposite direction, with these validation checks being implemented each in their own class. Rather than them all using one super-generic class. 
i.e. I would rather have several simple, dedicated classes rather than one abstract, complex class (this approach is complex, but the complexity is hidden in the simpleeval library)

That can be a future refactor, for now just implement this in the same style as the other checks. (i.e. don't use simpleeval)

The other reasons I'm against this
* The python code in the "formula" is exempt from our linting/typechecking
* It adds another tool (simpleeval) that someone maintaining this has to know
* The error messages are not simple and user-friendly

abis_mapping/templates/survey_site_data_v2/mapping.py
joecrowleygaia: consistency. Idea was to return map types from extractor methods. Specifically chose `dict[str, True]` here, since it is more specific than `dict[str, bool]` while being compatible by any consumer that expects a boolean.

joecrowleygaia:
Ok will change but I feel some of the concerns may need to be elaborated to maybe provide future support for it:
- Complexity - this is basically the same as the existing Frictionless row-constraint check built-in, which i don't see as complex, just powerful. It also has the benefit of being [documented](https://framework.frictionlessdata.io/docs/checks/row.html#row-constraint)
- Simpleeval - Is already a dependency, hence only changing the `poetry.lock` content-hash in this PR. It is supposedly a reduced set of instructions, and "safe", in contrast to the `eval` builtin interpreter. 
- Linting/Typechecking - This is a single line boolean expression, where we would be type hinting/asserting in any case since it is parsed data being used, which is itself potentially arbitrary i.e. the runtime types encountered can be different to the annotations we assign.
- Error messages - Incorporating the same as row-constraint at the moment. 

The only concern i had about this approach is the use of simpleeval and security. While it is stated it is "safe" i still don't feel 100% comfortable using evals, much the same reason i don't like seeing raw SPARQL, leaving the door ajar for potential injection attacks. Hence i had a felt need to limit the schema field names as much as possible and the like, being a source of potential integrity issues later on. For this alone i was hesitant, and along with your raised concerns, the reason I'll go back to doing what we have been doing.

PR #349, opened 2024-11-28
BDRSPS-1068 Add custom string field to treat whitespace-only cells as empty

Lincoln-GR:
This custom field is set to override the default frictionless string field.
It converts whitespace-only cells to the empty string before the regular frictionless validation/checks occur.

This means that
1. a required field will error when supplied with a whitespace-only cell.
2. All strings fields will appear an `None` when mapping, when their cell is whitespace-only (i.e. behave the same as an empty cell)

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,3 +92,15 @@ python docs/tables/vocabs.py <template_id> [-o] <output_file>
 where `output_file` (optional) corresponds to a location to store the resulting csv. 
 Default output is standard out
 if no `output_file` argument provided.
+
+## Frictionless
+
+The project uses the [Frictionless Framework](https://framework.frictionlessdata.io/) to validate and read the
+template CSV files.
+We have a number of plugins defined in the [`/abis_mapping/plugins`](/abis_mapping/plugins) directory which
+customize and extend how Frictionless works.
+
+### Overridden String Field Class
+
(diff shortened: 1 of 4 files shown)


joecrowleygaia:
LGTM. Should update `README.md` maybe since it is changing how frictionless `string` schema field behaves. Anyone that's new to project might get caught out chasing things when `string` ain't doing what they expect it do from the frictionless docs.

joecrowleygaia (commented):
comment left

README.md
```diff
@@ -92,3 +92,15 @@ python docs/tables/vocabs.py <template_id> [-o] <output_file>
...
+
+The project uses the [Frictionless Framework](https://framework.frictionlessdata.io/) to validate and read the
+template CSV files.
+We have a number of plugins defined in the [`/abis_mapping/plugins`](/abis_mapping/plugins) directory which
+customize and extend how Frictionless works.
+
+### Overridden String Field Class
+
+In particular, we have the [`abis_mapping/plugins/string_customized.py`](/abis_mapping/plugins/string_customized.py)
+plugin which overrides the default Frictionless string field to use our custom class.
```
joecrowleygaia, line 106: :+1:

joecrowleygaia (approved):

PR #352, opened 2024-11-29
BDRSPS-1084 Swap all void:inDataset predicates to schema:isPartOf

Lincoln-GR:
Every subject that has a `(..., void:inDataset, <dataset>)` triple, swap the `void:inDataset` predicate for `schema:isPartOf`


chungvl (approved):

chungvl:
Just the merge conflicts

PR #358, opened 2024-12-03
BDRSPS-1008 Add instructions for template file name requirements

Lincoln-GR:
Example of new docs:

image


serge-gaia (approved):

serge-gaia:
I'm thinking that we might want to explain a bit what we consider a manual submission for people that never used it.
Something like:
When making a manual submission through the web form ....
What you think @Lincoln-GR?

Lincoln-GR:
> When making a manual submission through the web form ....

@serge-gaia Yeah we could do that, I guess I was trying to not say too much about part-a since this is abis-mapping?

Lincoln-GR:
The website home page says "manual submission website"

image

So on second thoughts I reckon lets just go with this and see if we get any feedback

PR #363, opened 2024-12-09
BDRSPS-1118 Refactor `apply_mapping` method by moving it to the base mapper class

Lincoln-GR:
Each mapper class previously had very similar versions of the `apply_mapping` method. 
This has been refactored to a single method on the base class.
Now `apply_mapping_row` is the abstract method each subclass must implement


chungvl (approved):

serge-gaia:
Less code, DRY principle applied, and tests are passing. I'm happy :+1:

PR #374, opened 2024-12-20
BDRSPS-1133 Update IRI patterns for Agents

Lincoln-GR:
Use either `org/{name}` or `person/{name}` depending on which sort of agent it is.


abis_mapping/utils/iri_patterns.py
```diff
@@ -164,20 +164,22 @@ def datatype_iri(
 
 
 def agent_iri(
-    org: str,
+    agent_type: Literal["org", "person"],
```
serge-gaia, line 167: :+1:

serge-gaia (approved):
Looks good

PR #377, opened 2025-01-03
BDRSPS-1122 Make mapper `metadata` method return a frozen class instead of a dict

Lincoln-GR:
This is important since this method is cached, we don't want to the value stored in the cache to be mutated, which would effect future calls to the function.

### DIFF
--- a/abis_mapping/base/mapper.py
+++ b/abis_mapping/base/mapper.py
@@ -335,7 +335,7 @@ class ABISMapper(abc.ABC):
 
         # Template File is the name and filetype as extension from metadata
         md = cls.metadata()
-        template_file = directory / f"{md['name']}.{md['file_type'].lower()}"
+        template_file = directory / f"{md.name}.{md.file_type.lower()}"
 
         # Return
         return template_file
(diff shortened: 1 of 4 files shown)


abis_mapping/base/mapper.py
```diff
@@ -201,25 +200,6 @@ def add_geometry_supplied_as(
...
-        directory = pathlib.Path(inspect.getfile(cls)).parent
-        schema_file = directory / "schema.json"
-
-        # Get raw schema
-        fields: list[dict[str, object]] = json.loads(schema_file.read_text())["fields"]
-        out_path = directory / f"{cls.metadata()['name']}.{cls.metadata()['file_type'].lower()}"
-        with out_path.open("w") as f:
-            csv_writer = csv.DictWriter(f, [field["name"] for field in fields])
-            csv_writer.writeheader()
-
```
serge-gaia, line 222: unused by partA I guess.

abis_mapping/base/mapper.py
```diff
@@ -367,30 +347,31 @@ def template_id(self) -> str:
...
         """
         # Retrieve Metadata Filepath
         directory = pathlib.Path(inspect.getfile(cls)).parent
         metadata_file = directory / "metadata.json"
 
         # Read Metadata and validate
-        md_dict = json.loads(metadata_file.read_text())
-        md_class = models.metadata.TemplateMetadata.model_validate(md_dict, strict=False)
+        md_content = metadata_file.read_bytes()
+        md_class = models.metadata.TemplateMetadata.model_validate_json(md_content)
```
serge-gaia, line 367: Why the `strict=False` is gone? I kind of remember there was a reason for putting it to `False`

abis_mapping/base/mapper.py
Lincoln-GR: yep

abis_mapping/base/mapper.py
serge-gaia: Ok, I think I see. Is it because we pass bytes now, so no coertion is done?

Lincoln-GR: Before we were using `model_validate` which was being passed a python dict of the metadata (result of json.loads). 
Because the model has an enum field the `strict=False` was required so it would accept a string for the enum field.
(instead of an actual enum instance)

I think because we are now using `model_validate_json` instead which is passed a JSON string,
(i.e. pydantic handles the json-decoding)
pydantic knows that can never have an actual enum instance, and it turns the strings into the enum without needing `strict=False`

serge-gaia (approved):
Approving but see comment about the removal of strict mode in `models.metadata.TemplateMetadata.model_validate_json(md_content)`

abis_mapping/base/mapper.py
serge-gaia: Easy to test with partA API

PR #384, opened 2025-01-30
[BDRSPS-1173]

chungvl:


abis_mapping/templates/survey_site_data_v3/mapping.py
```diff
@@ -669,6 +672,8 @@ def add_habitat_collection(
         graph.add((uri, rdflib.SDO.member, site))
         # Add link to attribute
         graph.add((uri, utils.namespaces.TERN.hasAttribute, attribute))
+        if submission_iri:
+            graph.add((uri, rdflib.VOID.inDataset, submission_iri))
```
chungvl, line 677: Hi - @Lincoln-GR  like this?

Lincoln-GR: yep looks good :+1:

Lincoln-GR (commented):
Looks good so far :+1:, but there are more places in this template that need the new triple added.

Also, can you add `submission_iri` to the args in the docstrings of the changed functions. This is to follow our code conventions.

Lincoln-GR:
also, another thought, I think it would be best to leave out the `= None` default values everywhere for the new argument. That way if we forget to pass it anywhere we'll get an error, rather than it silently defaulting to None.

PR #387, opened 2025-02-03
BDRSPS-1173

chungvl:


abis_mapping/templates/survey_site_data_v3/mapping.py
```diff
@@ -600,6 +606,8 @@ def add_habitat_attribute(
         """
         # Add type
         graph.add((uri, a, utils.namespaces.TERN.Attribute))
```
chungvl, line 608: @Lincoln-GR  am I correct in inferring that `TERN.Attribute` is the only relevant node in this mapping?

Lincoln-GR: There are also `rdflib.SDO.Collection`s (`schema:Collection`)

abis_mapping/templates/survey_site_data_v3/mapping.py
```diff
@@ -669,6 +683,8 @@ def add_habitat_collection(
         graph.add((uri, rdflib.SDO.member, site))
         # Add link to attribute
         graph.add((uri, utils.namespaces.TERN.hasAttribute, attribute))
+        if submission_iri:
+            graph.add((uri, rdflib.VOID.inDataset, submission_iri))
```
Lincoln-GR, line 687: This is duplicating the change above

Lincoln-GR (commented):
Looks good @chungvl, one small thing, plus I think we should make the tests exercise the new mapping.

Lincoln-GR (approved):
Nice! Now we can see what the updated RDF looks like

PR #393, opened 2025-02-04
BDRSPS-1173: `survey occurrence v3`

chungvl:


Lincoln-GR (changes requested):
Nice start, there are few places missed I think

* The `TERN.SiteVisit` needs a link
* some of the `TERN.FeatureOfInterest` still need links

Plus the `TERN.Sampling`s need links. Note this was just added to ticket after I clarified with Mieke, apologies.

Lincoln-GR (approved):
Thanks for adding those missing ones :+1:

PR #399, opened 2025-02-06
BDRSPS-1192

chungvl:


abis_mapping/templates/survey_metadata_v3/mapping.py
```diff
@@ -645,6 +644,17 @@ def add_plan(
             dataset: URI for the dataset node.
             graph: Graph to be modified.
         """
+
```
Lincoln-GR, line 647: It would be good to have an explanatory comment here

Lincoln-GR (changes requested):

Lincoln-GR (approved):

PR #403, opened 2025-02-07
BDRSPS-1203 Fix slugify-ing slashes in values

Lincoln-GR:
Add new `uri_slugified()` function and use it everywhere appropriate, so that slashes (`/`) in field values are replaced with dashes when being slugified.

The old function `uri()` purposely kept slashes, this was intended for keeping slashes **between** field values, but unintentionally also kept slashes **in** field values.

Since `uri()` is now only ever called with static values, its slugifying has been removed, since all static values passed to it are already slugified.

Also did a improvement so `agent_iri()` is only called when there is an actual value in the field being used for the IRI.

### DIFF
--- a/abis_mapping/utils/iri_patterns.py
+++ b/abis_mapping/utils/iri_patterns.py
@@ -124,7 +124,7 @@ def attribute_iri(
     Returns:
         IRI for the tern:Attribute node.
     """
-    return utils.rdf.uri(f"attribute/{attribute}/{value}", namespace=base_iri)
+    return utils.rdf.uri_slugified(base_iri, "attribute/{attribute}/{value}", attribute=attribute, value=value)
 
 
 def attribute_value_iri(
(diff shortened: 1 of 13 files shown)


chungvl (approved):

abis_mapping/utils/iri_patterns.py
```diff
@@ -168,7 +164,13 @@ def attribute_collection_iri(
     Returns:
         IRI for the schema:Collection node.
     """
-    return utils.rdf.uri(f"{collection_type}Collection/{attribute}/{value}", namespace=base_iri)
+    return utils.rdf.uri_slugified(
+        base_iri,
+        "{collection_type}Collection/{attribute}/{value}",
```
serge-gaia, line 169: the lack of 'f` is slighly disturbing at first :-)

PR #407, opened 2025-02-10
[BDRSPS-1209] Remove creation of `hasPlan` if no plan

chungvl:


abis_mapping/templates/survey_metadata_v3/mapping.py
```diff
@@ -247,10 +247,18 @@ def apply_mapping_row(
...
 
+        # Add plan
+        plan = self.add_plan(
+            uri=survey_plan,
+            row=row,
+            dataset=dataset,
+            graph=graph,
+        )
+
         # Add BDR survey
```
chungvl, line 258: @Lincoln-GR  is this on the right track?

Lincoln-GR: I think this would work, but I wouldn't do it this way because none of the other mapping code uses the "return bool" pattern. We have a lot of mapping code, keeping it consistent is good.

Instead we have been following the pattern of "If the URI shouldn't be in the graph, then make the variable holding the URI `None`", i.e. don't construct the URI in the first place.

So in this case I would move the "if fields have values" check out of `add_plan` and put it before setting the `survey_plan` variable. e.g.

```python
if (
    row["targetTaxonomicScope"]
    or ...  # etc
):
    survey_plan = utils.iri_patterns.plan_iri(...)
else:
    survey_plan = None
```

That way you should get type or runtime errors if the URI is being added to the graph when it shouldn't.

abis_mapping/templates/survey_metadata_v3/mapping.py
```diff
@@ -637,7 +648,7 @@ def add_plan(
         row: frictionless.Row,
         dataset: rdflib.URIRef,
         graph: rdflib.Graph,
-    ) -> None:
+    ) -> bool:
```
Lincoln-GR, line 651: let's revert this back to `-> None`

Lincoln-GR (commented):
Looks good :+1: Few more mypy errors to fix

Lincoln-GR (approved):
Nice work! :rocket:

PR #456, opened 2025-05-22
[BDRSPS-1463] Submission instructions change as per Kats instructions

chungvl:


Lincoln-GR (changes requested):
Page title needs to be updated as well

docs/pages/submission_properties.md
```diff
@@ -2,15 +2,14 @@
 title: Submission Properties Form
```
Lincoln-GR, line 2: @chungvl This line is the title of the page in the documentation navigation panel. It should be updated as well

Lincoln-GR (changes requested):

docs/pages/submission_properties.md
serge-gaia: That was me :-)

chungvl: Ah

